PR #3, opened 2019-09-13
pod: Support `runAsUser: 0`

cgwalters:
I think this is better than calling it `privileged` as that
has a specific meaning in Docker/Kubernetes.

See my argument at https://lwn.net/Articles/796885/


cgwalters:
(Not tested)

PR #4, opened 2019-09-13
pod: fix map lookup for `runAsUser: 0`

jlebon:
This is the classic "handle 'falsy' values in dict lookup" problem. We
need to compare against `null` to actually check for existence in the
map in case the value is `0`.


cgwalters:
LGTM

PR #6, opened 2019-12-06
pod: add support for cpu requests

jlebon:
Going to use this in rpm-ostree CI.

The `.toString()` there is necessary. Otherwise, Groovy gets super
confused and interprets the values as an array of different types.:

```
resources:
  requests:
    cpu:
      bytes:
      - 54
      strings:
      - ''
      - ''
      valueCount: 1
      values:
      - 6
```

Possibly Groovy's loose type system gone wrong?


jlebon:
Just gonna self-merge this. Tested working in the rpm-ostree CI.

PR #8, opened 2020-02-06
coreos.groovy: always mount emptyDir on /srv

jlebon:
Let's designate `/srv` as the place where CI work that needs a real
filesystem and does lots of IO can operate.

### DIFF
--- a/vars/coreos.groovy
+++ b/vars/coreos.groovy
@@ -33,13 +33,19 @@ def pod(params, body) {
     if (params['cpu']) {
         podObj['spec']['containers'][1]['resources']['requests']['cpu'] = params['cpu'].toString()
     }
-    if (params['emptyDirs']) {
-        podObj['spec']['volumes'] = []
-        podObj['spec']['containers'][1]['volumeMounts'] = []
-        params['emptyDirs'].eachWithIndex { mountPath, i ->
-            podObj['spec']['volumes'] += ['name': "emptydir-${i}".toString(), 'emptyDir': [:]]
-            podObj['spec']['containers'][1]['volumeMounts'] += ['name': "emptydir-${i}".toString(), 'mountPath': mountPath]
-        }
+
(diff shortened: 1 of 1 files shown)


jlebon:
Hmm actually, I'm confused now how the cosa's CI today isn't hitting https://github.com/coreos/coreos-assembler/blob/ec5c0ee036d91f1f417c7fac7c28804cdcde9324/src/cmdlib.sh#L119. Ahh, unless CentOS CI is using the device mapper backend? Let me sanity check that before merging this.

jlebon:
Ahh yes, it is indeed on devmapper:

```
# df -hT /
Filesystem                                                                                          Type  Size  Used Avail Use% Mounted on
/dev/mapper/docker-253:0-402693158-390ffb0b21b35da3751d93fb52701d893dfecfc4574c74632dfd458b3a376908 xfs    20G  240M   20G   2% /
```

But e.g. we're allocated 20G, which yeah... seems like it's possible we could go over in cosa with all the buildextends we do?

With emptyDir:

```
# df -hT / /srv
Filesystem                                                                                          Type  Size  Used Avail Use% Mounted on
/dev/mapper/docker-253:0-134218338-4b500582577a1a5875d397b57b2d141d7ec615b684a1bb09aaddb4a9f724cbf8 xfs    20G  240M   20G   2% /
/dev/mapper/vg_n27-root                                                                             xfs   221G   46G  176G  21% /srv
```

> Seems sane to me, though I wonder if this will result in more pod evictions today. I know we were seeing that with RHCOS at one point.

Yeah, it's not clear to me whether there's quota restrictions on emptyDirs. The [documentation](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) doesn't mention it. I found https://github.com/kubernetes/kubernetes/issues/13479 which is related, though I'm not sure if the quota on emptyDir was ever implemented.

Anyway, let's try this out and see how it goes!

PR #9, opened 2020-02-13
coreos.groovy: add tmp hack to avoid papr SA on CoreOS CI

jlebon:
We'll be able to drop this entirely once we've moved over to CoreOS CI.
But until then, let's require code running on CoreOS CI to pass
`coreosci: true` and key off of that to determine if to use the papr
SA.

I wanted to base this off of the `JENKINS_URL` environment variable at
first, but I couldn't get it to work.


jlebon:
Nevermind this, I think we'll just do a hard cutover when we're ready to migrate.

PR #10, opened 2020-02-13
Add more custom steps, streamline cosa build + kola workflow

jlebon:
This changes in significant ways how this shared library is set up.
First, I re-read the documentation on shared libraries more closely and
we now follow conventions like having custom steps use a `call()`
function.

But also, there is now a much bigger emphasis on the `cosa build`
workflow. For example, there is now a `cosaPod` which default to the
cosa image.

There's a `buildroot` parameter one can use to request the buildroot
image instead (and of course, there's a lower-level `pod` step for full
control).

There is also a `fcosBuild` custom step, which will automatically build
FCOS, optionally building and/or layering some directories, then run
`cosa kola`.

### DIFF
--- a/vars/cosaPod.groovy
+++ b/vars/cosaPod.groovy
@@ -1,0 +1,15 @@
+// Thin wrapper around pod() which auto-selects images and kvm.
+// Additional available parameters:
+//   image: string
+//   buildroot: bool
+def call(params = [:], Closure body) {
+    if (params['image'] == null) {
+        if (params['buildroot']) {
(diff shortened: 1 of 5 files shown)


jlebon:
This will not work with projectatomic-ci, so requires https://github.com/jlebon/coreos-ci/pull/1.

To not break repos currently hooked up to projectatomic-ci, I'm targeting the `coreos-ci` branch. And once everything is migrated, we can merge it back into master.

jlebon:
I'm just going to merge this since I think we're still in the "foundation" phase where some broken code is OK and nothing is relying on it yet.

PR #11, opened 2020-02-20
fcosBuild: add skipInit parameter

jlebon:
Sometimes the procedure to set up the cosa workdir can be more complex
depending on what we're testing. Allow callers to skip the `cosa init`
implicitly done here.


miabbott:
How will users setup the workdir in these more complex scenarios?

jlebon:
It's basically any situation where you don't want the vanilla `cosa init https://github.com/coreos/fedora-coreos-config`. E.g. if you want to init from a local directory. See https://github.com/coreos/fedora-coreos-config/pull/285 for an example.

PR #12, opened 2020-03-06
Merge coreos ci branch

cgwalters:
We can do this now that the projectatomic-ci is dead!


jlebon:
Follow-up: https://github.com/coreos/coreos-ci/pull/11

PR #15, opened 2020-03-12
ci: hook up to CoreOS CI

jlebon:
And of course, we should test our test scripts themselves. So hook this
repo up to CoreOS CI.


cgwalters:
Hm, the CI system seems confused by this but it's not clear to me why.

jlebon:
Yeah, I need to dig more into this. Essentially, the problem is that Jenkins gets confused because we're asking it to load a shared library that's already loaded (because `coreos-ci-lib` is already in the global set of shlibs). Will have to figure out a hack around this.

jlebon:
I think this would still be good to fix eventually, but it's not high enough in priority. We've learned to just point the stg Jenkins to a fork of coreos-ci-lib to test things.

PR #16, opened 2020-03-19
fcosKola: Automatically pick up tests/kola 

cgwalters:
So we can use `-E` in CI, see
https://github.com/coreos/coreos-assembler/pull/1252

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -3,8 +3,12 @@ def call(cosaDir = "/srv/fcos") {
     stage('Kola') {
         parallel run: {
             stage("run") {
+                def args = ""
+                if (shwrapRc("test -d tests/kola") == 0) {
+                    args += "--exttest ${env.WORKSPACE}/tests/kola"
+                }
                 try {
-                    shwrap("cd ${cosaDir} && cosa kola run --parallel 8")
+                    shwrap("cd ${cosaDir} && cosa kola run --parallel 8 ${args}")
                 } finally {
                     shwrap("tar -c -C ${cosaDir}/tmp kola | xz -c9 > ${env.WORKSPACE}/kola.tar.xz")
                     archiveArtifacts allowEmptyArchive: true, artifacts: 'kola.tar.xz'


cgwalters:
(Totally didn't test this)

jlebon:
Ahh yup, you're right we need to enhance this step here for https://github.com/coreos/fedora-coreos-config/pull/305.

Hmm, how about instead of this, we automatically pass `--exttest src/config/tests/kola` if we find that the directory exists? In general, I want to establish conventions to make it dead easy to do the right thing.

jlebon:
I can pick this up if you'd like!

cgwalters:
>  we automatically pass --exttest src/config/tests/kola

Oh definitely we should do that yep.  But this approach is needed for the ostree case...unless we define conventions for that.

jlebon:
Yeah, I think that makes sense. I.e. let's define `tests/kola` as the external tests dir. For config dirs like fedora-coreos-config, we check for this via `${cosaDir}/src/config/tests/kola`. For the project being built, we just check `tests/kola` (i.e. the workspace). This matches e.g. `make: true` blindly building the code in the workspace.

For things that break that convention or for tests stored elsewhere, we can add an `exttests: []` arg that override the default list? But that can come later.

cgwalters:
Oh I like that yep!

cgwalters:
https://github.com/coreos/coreos-assembler/pull/1261

cgwalters:
And reworked this per above discussion.

vars/fcosKola.groovy
```diff
@@ -3,8 +3,13 @@ def call(cosaDir = "/srv/fcos") {
     stage('Kola') {
         parallel run: {
             stage("run") {
+                def args = ""
+                def srctests = new File("tests/kola").getAbsolutePath()
+                if (srctests.isDirectory()) {
```
jlebon: This unfortunately doesn't work in Kubernetes pods, e.g. the pipeline has: https://github.com/coreos/fedora-coreos-pipeline/blob/2c6c49599fd242e3ffbc6d4b4b1087fdb731a5ae/utils.groovy#L46. Which, yeah OK let me upstream the `shwrap_rc` function to this repo: https://github.com/coreos/coreos-ci-lib/pull/17. Then we can use it here for this.

I'll double check whether this is fixed, and if not, also upstream `path_exists` as a helper.

cgwalters: Ah, nice catch.  OK I took another stab at this but still not tested.

jlebon:
I just did a minor fixup to this to avoid another round-trip. (Changed `WORKSPACE` to `env.WORKSPACE` and just spelled out the `--exttest` switch to be more explicit.)

Thanks!

cgwalters:
I pushed https://github.com/coreos/coreos-ci-lib/commit/55bb2ae8e693beea2249dfc37e8e6968c0b978c5

PR #17, opened 2020-03-19
Add shwrapRc helper function

jlebon:
So you can do things like `test -s foobar` more easily.
Upstreamed from fedora-coreos-pipeline.


jlebon:
I changed this to camel-case to go with the convention.

PR #18, opened 2020-03-23
fcosKola: Avoid duplicate registration of tests/kola

cgwalters:
This started failing in the fedora-coreos-config CI.

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -4,7 +4,14 @@ def call(cosaDir = "/srv/fcos") {
         parallel run: {
             stage("run") {
                 def args = ""
-                if (shwrapRc("test -d tests/kola") == 0) {
+                // Add the tests/kola directory, but only if it's not the same as the
+                // src/config repo which is also automatically added.
+                if (shwrapRc("""
+                    test -d tests/kola
+                    configorigin=\$(cd ${cosaDir}/src/config & git config --get remote.origin.url)
+                    gitorigin=\$(cd ${env.WORKSPACE} && git config --get remote.origin.url)
+                    test "\$configorigin" -ne "\$gitorigin"
+                """) == 0) {
                     args += "--exttest ${env.WORKSPACE}"
                 }
                 try {


cgwalters:
(Totally untested as I don't have a workflow for that)

vars/fcosKola.groovy
```diff
@@ -4,7 +4,14 @@ def call(cosaDir = "/srv/fcos") {
         parallel run: {
             stage("run") {
                 def args = ""
-                if (shwrapRc("test -d tests/kola") == 0) {
+                // Add the tests/kola directory, but only if it's not the same as the
+                // src/config repo which is also automatically added.
+                if (shwrapRc("""
+                    test -d tests/kola
+                    configorigin=$(cd tests/kola & git config --get remote.origin.url)
```
jlebon: Hmm, I think this should be `cd ${cosaDir}/src/config`, right?

vars/fcosKola.groovy
```diff
@@ -4,7 +4,14 @@ def call(cosaDir = "/srv/fcos") {
...
+                    gitorigin=$(cd ${env.WORKSPACE} && git config --get remote.origin.url)
+                    test "$configorigin" -ne "$gitorigin"
```
jlebon: This is an easy shell-in-Groovy gotcha, but we need to escape all the `$` which we want literally in bash. So here, `test "\$configorigin" ...`. Except `${env.WORKSPACE}` since that's a Groovy substitution.

jlebon:
> (Totally untested as I don't have a workflow for that)

See https://github.com/coreos/coreos-ci/pull/14, but we don't have to block on that.

vars/fcosKola.groovy
cgwalters: Yep, good catch.

cgwalters:
OK [this is broken](https://jenkins-coreos-ci.apps.ci.centos.org/blue/organizations/jenkins/github-ci%2Fostreedev%2Fostree/detail/PR-1995/9/pipeline/149):

```
/home/jenkins/agent/workspace/ci_ostreedev_ostree_PR-1995-HQCNGX3O6J7HRHAGUUNPR4SO5JIYXILGHBQDC2VAGSSUBUNPBIPQ@tmp/durable-11b4f6c7/script.sh: line 7: test: https://github.com/ostreedev/ostree.git: integer expression expected
```

cgwalters:
https://github.com/coreos/coreos-ci-lib/commit/94b20a88291dc1cea932cf170352414a71361311

PR #19, opened 2020-04-07
fcosKola: allow skipping upgrade tests

jlebon:
In `next-devel` right now, we don't have a previous build yet, so
`run-upgrade` doesn't know what parent build to use. There's a good
argument here that we could use the `testing-devel` stream as the
starting stream for now (and actually, that's something that we should
enhance our CI to cover anyway, even when we *do* have `next-devel`
parent builds, since that mimics what nodes will eventually do).

But I just want to get `next-devel` off the ground for now. The first
step either way is allowing us to skip upgrade testing here so that we
can run it manually with more options as needed.

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -1,30 +1,44 @@
 // Run kola tests on the latest build in the cosa dir
-def call(cosaDir = "/srv/fcos") {
-    stage('Kola') {
-        parallel run: {
-            stage("run") {
-                def args = ""
-                // Add the tests/kola directory, but only if it's not the same as the
-                // src/config repo which is also automatically added.
-                if (shwrapRc("""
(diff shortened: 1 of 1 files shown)


jlebon:
I think I got the syntax right. If that's not the case, I'll push up fixes as needed!

jlebon:
Required by https://github.com/coreos/fedora-coreos-config/pull/306.

cgwalters:
This broke ostree's CI at least https://github.com/ostreedev/ostree/pull/2068#issuecomment-614142863

jlebon:
https://github.com/ostreedev/ostree/pull/2069

cgwalters:
If the "new" jenkins UI didn't silently swallow stack traces this would have been way faster for me to figure out...

PR #20, opened 2020-04-15
fcosKola: Add tests parameter

cgwalters:
So we can choose to run just e.g. `ext.*` in CI runs for projects
to run just their own tests.

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -1,7 +1,9 @@
 // Run kola tests on the latest build in the cosa dir
 // Available parameters:
 //    cosaDir:         string  -- cosa working directory
+//    parallel:        integer -- Default 8 (this may be changed in the future)
 //    skipUpgrade:     boolean -- skip running `cosa kola --upgrades`
+//    extraArgs:       string  -- kola args (usually a glob pattern like `ext.*`)
 def call(params = [:]) {
     def cosaDir = "/srv/fcos"
     if (params['cosaDir']) {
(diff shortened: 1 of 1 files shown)


cgwalters:
(Completely untested)

cgwalters:
Or maybe this is really just `kolaArgs` since it's not like we're checking they're not just passing `--whatever`.

jlebon:
> Or maybe this is really just `kolaArgs` since it's not like we're checking they're not just passing `--whatever`.

Yeah makes sense. Or maybe just `extraArgs` since it's understood from the name of the step that it's for kola?

(If we wanted to special-cased for tests, I think we'd want it as a []string that we manually join instead?)

cgwalters:
OK some experimentation in [a groovy shell](https://groovy-playground.appspot.com/) gives me some confidence in this.

cgwalters:
Oh god no it'll be `null`

cgwalters:
Man...this is where Rust's `Option<>` and combinators like `.or_else()` are just SO MUCH BETTER.

vars/fcosKola.groovy
```diff
@@ -27,8 +28,12 @@ def call(params = [:]) {
             {
                 args += "--exttest ${env.WORKSPACE}"
             }
+            def extraArgs = params['extraArgs'];
+            if (extraArgs == null) {
+                extraArgs = ""
+            }
             try {
-                shwrap("cd ${cosaDir} && cosa kola run --parallel 8 ${args}")
+                shwrap("cd ${cosaDir} && cosa kola run --parallel 8 ${args} ${extraArgs}")
```
jlebon, line 36: Actually, should we also add a `parallel` argument while we're here? That way the rpm-ostree CI can set it to something it accounted for when calculating pod resource requests.

cgwalters: OK done.  I also found the two-argument `.get()` method which cleans up default handling.

That said...I think in the future we should teach kola how to auto-calculate this itself.

PR #21, opened 2020-04-16
fcosKola: Fix check for tests/kola

cgwalters:
rpm-ostree still isn't running the external tests, I think
we need this.

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -21,7 +21,7 @@ def call(params = [:]) {
             // Add the tests/kola directory, but only if it's not the same as the
             // src/config repo which is also automatically added.
             if (shwrapRc("""
-                test -d tests/kola
+                test -d ${env.WORKSPACE}/tests/kola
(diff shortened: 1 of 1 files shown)


jlebon:
Or wait, hmmm... actually. Let me think on this a bit more.

vars/fcosKola.groovy
```diff
@@ -21,7 +21,7 @@ def call(params = [:]) {
             // Add the tests/kola directory, but only if it's not the same as the
             // src/config repo which is also automatically added.
             if (shwrapRc("""
-                test -d tests/kola
+                test -d ${cosaDir}/tests/kola
```
jlebon, line 24: `cosaDir` is where we do cosa builds, but not necessarily where the project's git repo is checked out, right? This code is assuming that we're in `WORKSPACE`, so maybe what we really need is

```suggestion
                test -d ${env.WORKSPACE}/tests/kola
```

?

cgwalters: Oh yes...done.

PR #23, opened 2020-04-20
fcosBuild: use `cosa build --strict`

jlebon:
We always want to build in strict mode for CI. See:
https://github.com/coreos/fedora-coreos-tracker/issues/454


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1370

PR #26, opened 2020-04-27
fcosBuild: add extraFetchArgs

jlebon:
Similar to `extraArgs` but for `cosa fetch`. Sadly, building and testing
software is complex and there's always going to be corner cases that
burst through our nice and clean abstractions.


cgwalters:
Perhaps later split this up into
`fcosInit()` `fcosFetch()` `fcosBuildInitial()` or something?

PR #32, opened 2020-07-09
pod: drop hardcoded jnlp container

jlebon:
Instead, let the Kubernetes Jenkins plugin inject the default jnlp
container to use so that we're sure it's always using one that's made
for the given k8s version and plugin version.

This is causing issues on the new cluster where the `sh` step would
hang. It seems to be related to running on new k8s but old Jenkins, and
some kind of conflict with how working directories have changed. For
more details, see:

https://issues.jenkins-ci.org/browse/JENKINS-58777
https://issues.jenkins-ci.org/browse/JENKINS-58766

### DIFF
--- a/resources/com/github/coreos/pod.json
+++ b/resources/com/github/coreos/pod.json
@@ -4,14 +4,6 @@
     "spec": {
         "containers": [
             {
-                "name": "jnlp",
-                "image": "jenkins-slave-base-centos7:latest",
-                "args": [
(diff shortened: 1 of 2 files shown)


jlebon:
Re. old Jenkins, I tried updating to the latest OCP4-bundled one but hit some sort of compat issues with the GitHub OAuth plugin. I'd like to dig into those in a separate namespace though for now, I just want to focus on having CoreOS CI be solid again.

PR #33, opened 2020-07-16
cosaPod: spawn container with dumb-init

jlebon:
Let's use `dumb-init` by default for cosa pods so that we don't end up
with a swarm of zombies taking up resources.

I have a feeling this might also be contributing to the ulimit
exhaustion we're seeing in CI after the move to the new OCP 4.4 cluster.

See also https://github.com/kubernetes/kubernetes/issues/84210.

### DIFF
--- a/vars/cosaPod.groovy
+++ b/vars/cosaPod.groovy
@@ -16,6 +16,9 @@ def call(params = [:], Closure body) {
         params['kvm'] = true
     }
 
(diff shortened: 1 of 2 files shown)


cgwalters:
Ahhh man.  Of course.   It's so insane that *years* after Docker was invented we are *still* dealing with this tech debt.

LGTM!

PR #34, opened 2020-08-05
shwrap: Fix umask if it's not set

cgwalters:
We never want a 0 umask; it can e.g. cause world writable directories
which triggers failures in e.g. ostree's CI tests.


cgwalters:
xref https://github.com/ostreedev/ostree/pull/2163

jlebon:
OK, narrowed this down to a specific node at least! :tada:
Filed: https://pagure.io/centos-infra/issue/48

Once we get to the bottom of this, let's revert this commit.

PR #36, opened 2020-08-26
Add shwrapCapture() step

jlebon:
This is a variant on shwrap() which returns the output of the command.
Part of draining helper functions from the pipeline into the library in
prep for modeling the pipeline jobs after upstream CI.


miabbott:
LGTM

PR #37, opened 2020-08-26
pod: add support for mounting secrets

jlebon:
Part of the series on standardizing on coreos-ci-lib in production. We
have this intermediate world right now where all the secrets are being
mounted into all pod invocations, which doesn't really make sense. As an
intermediate step, let's allow mounting only the secrets we actually
need into the pod.

Eventually though we should use native Jenkins credentials so that we
only have access to the secrets for the specific steps we need.


miabbott:
LGTM

jlebon:
Thanks Micah for the reviews!

PR #39, opened 2020-09-25
pod: automatically generate env vars for secret paths

jlebon:
That way the closure doesn't have to create its own Groovy variable, or
rewrite the full path multiple times.

The substitution is pretty simple: we swap all dashes to underscores,
and make it all uppercase. So `aws-fcos-builds-bot-config` would become
`AWS_FCOS_BUILDS_BOT_CONFIG`.


dustymabe (approved):
LGTM

PR #42, opened 2020-10-27
Improve coreos-ci-lib

ravanelli:
- Add utils with common functions used in CI
- Add cosaDir, skipDirCreation and changeOwner
parameters in fcosBuild
- Add self test and and basic scenarios tests in fcosKola

### DIFF
--- a/vars/fcosBuild.groovy
+++ b/vars/fcosBuild.groovy
@@ -7,34 +7,46 @@
 //    overlays:     []string  -- list of directories to overlay
 //    extraFetchArgs: string  -- extra arguments to pass to `cosa fetch`
 //    extraArgs:      string  -- extra arguments to pass to `cosa build`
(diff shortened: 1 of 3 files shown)


jlebon:
Hi @ravanelli, thanks for the patch! Can you provide more context on this?

Also, do you mind if we split this out over multiple PRs? It'll make it easier to review. The `cosaDir` support in `fcosBuild` looks like a good start. :)

ravanelli:
Hi @jlebon,  thanks for checking it.
The idea of this patch is to improve the coreos-ci-lib in a way we can really use the lib across all CIs. At this moment we have duplicate code all over the different CIs:  FCOS, COSA, RHCOS and so. Would be good to have this duplicity cut off, making the CI easier to maintain and use.
Sure! I will split it :)

PR #43, opened 2020-10-28
fcosBuild: Add `cosaDir`parameter and getCosaDir

ravanelli:
- Add `cosaDir` parameter to be used across multiples CIs
making it usefull to change cosa working directory
- Add getCosaDir in utils as a place with common
functions used in CI

### DIFF
--- a/vars/utils.groovy
+++ b/vars/utils.groovy
@@ -1,0 +1,7 @@
+def getCosaDir(params = [:]) {
+    if (params['cosaDir']) {
+        return params['cosaDir']
+    } else {
+        return "/srv/fcos"
+    }
+}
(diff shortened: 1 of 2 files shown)


vars/utils.groovy
```diff
@@ -0,0 +1,8 @@
+def getCosaDir(params = [:]) {
+
```
jlebon, line 2: Minor: extra space here.

jlebon (commented):
One super minor/optional comment, LGTM otherwise!
Were you able to test this?

vars/utils.groovy
ravanelli: Fixed.
Yes, I tested it in my centos namespace using the setup for cosa CI

jlebon (approved):
Thanks!

PR #45, opened 2020-10-28
utils: Add cosaBuild and unitTest

ravanelli:
Make the cosa build process with unit tests available
in utils

### DIFF
--- a/vars/utils.groovy
+++ b/vars/utils.groovy
@@ -1,7 +1,48 @@
+def cosaBuild(params = [:]) {
+// Available parameters:
+//     gitInstall:         boolean -- install git package
+//     gitClone:           boolean -- clone cosa
+//     skipUnitTest:       boolean -- do not run unit tests
+    stage("Build Cosa") {
+        if (params['gitInstall']) {
+            shwrap(""" dnf install -y git """)
+        }
+
+        if (params['gitClone']) {
+            shwrap("""
+                git clone https://github.com/coreos/coreos-assembler.git
+                cd coreos-assembler && git submodule update --init
+                ./build.sh
+            """)
+        } else {
+            shwrap("""
+                git submodule update --init  && ./build.sh
+            """)
+       }
+       shwrap("""
+           rpm -qa | sort > rpmdb.txt
+       """)
+       archiveArtifacts artifacts: 'rpmdb.txt'
+    }
+
+    if (!params['skipUnitTest']) {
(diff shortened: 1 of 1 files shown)


jlebon:
Hmm, I'm confused why these should belong here. Instructions to build and unit test cosa itself should live as part of the codebase in https://github.com/coreos/coreos-assembler.

darkmuggle:
@jlebon so you see CI lib as the place for running CI and the COSA commands should be inside COSA exclusively? Looking over COSA Lib there's plenty of specific things to COSA (i.e. kola here.) The line is a bit....blurred.

I *think* that have the ability to start COSA and import the libs to run COSA is fine.

jlebon:
Running `cosa` commands like `cosa kola`, `cosa build`, etc... is totally fine because they're "CLI APIs" for building and testing the target OS. But building and unit testing cosa itself is more tied to the codebase.

There's some APIs which we can use to abstract it away (see how for example `fcosBuild` supports doing `make && make install DESTDIR=...`) and I could imagine a `make()` step here if wanted to go that route (PAPR for example did this: https://github.com/projectatomic/papr/blob/master/docs/sample.papr.yml#L141-L160). But having business logic for building and testing a specific codebase seems out of place here. (I agree cosa is special for us though, but still... conceptually that knowledge belongs better alongside the code itself; for example, it naturally gets forked when we branch for OCP releases.)

Maybe another way to frame this: this shared lib isn't a "cosa library", it's a "CoreOS CI library". E.g. the `pod` step here totally supports using whatever image you want and not run cosa at all. Though of course, because we use cosa heavily across the projects, there's a lot of sugar for that path.

Hope that makes sense!

ravanelli:
> Running `cosa` commands like `cosa kola`, `cosa build`, etc... is totally fine because they're "CLI APIs" for building and testing the target OS. But building and unit testing cosa itself is more tied to the codebase.
> 
I think we also need to define what is business logic in this case. IHMO
The `make` and `test` are tied to the CI and not to the codebase, since cosa have it's own [build.sh](https://github.com/ravanelli/coreos-assembler/blob/master/build.sh) What the lib is doing is touching the [cci.jenkinsfile](https://github.com/ravanelli/coreos-assembler/blob/master/.cci.jenkinsfile) that is only related to the CI process, it will not change the business logic.

ravanelli:
I'm closing it since we decided to not add cosa logic here.

PR #47, opened 2020-10-30
pod: add support for configMaps

jlebon:
We'll need this in order to move from the old world (of hardcoding
everything in a `pod.json` file in the git repo) to the new world (where
we only mount what we actually need for a given pod).


dustymabe (approved):
LGTM

PR #48, opened 2020-11-03
utils: Add testLive

ravanelli:
Make testLive available in utils

### DIFF
--- a/vars/fcosKolaTestIso.groovy
+++ b/vars/fcosKolaTestIso.groovy
@@ -1,0 +1,46 @@
+// Run kola testiso
+// Available parameters:
+//     cosaDir:        string  -- cosa working directory
+//     extraArgs:      string  -- extra arguments to pass to `kola testiso`
+//     extraArgs4k:    string  -- extra arguments to pass to 4k `kola testiso`
+//     scenarios:      string  -- scenarios to pass to `kola testiso`
+//     scenarios4k:    string  -- scenarios to pass to `kola testiso`
+//     skipMetal4k:    boolean -- skip metal4k image
+def call(params = [:]) {
+    def cosaDir = utils.getCosaDir(params)
+    def extraArgs = params.get('extraArgs', "");
+    def extraArgs4k = params.get('extraArgs4k', "");
+    def scenarios = params.get('scenarios', "");
+    def scenarios4k = params.get('scenarios4k', "");
+
+    testIsoRuns = [:]
+    testIsoRuns["metal"] = {
+        try {
+            if (params['scenarios']) {
+                shwrap("cd ${cosaDir}  && kola testiso -S ${extraArgs} --scenarios ${scenarios} --output-dir tmp/kola-testiso-metal")
+            } else {
+                shwrap("cd ${cosaDir}  && kola testiso -S ${extraArgs} --output-dir tmp/kola-testiso-metal")
+            }
+        } finally {
+            shwrap("cd ${cosaDir} && tar -cf - tmp/kola-testiso-metal/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-metal.tar.xz")
+        }
+    }
+    if (!params['skipMetal4k']) {
+        testIsoRuns["metal4k"] = {
+            try {
+                if (params['scenarios4k']) {
+                    shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k ${extraArgs4k} --scenarios ${scenarios4k} --output-dir tmp/kola-testiso-metal4k")
+                } else {
+                    shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k ${extraArgs4k} --output-dir tmp/kola-testiso-metal4k")
+                }
+            } finally {
+                shwrap("cd ${cosaDir} && tar -cf - tmp/kola-testiso-metal4k/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-metal4k.tar.xz")
+            }
+        }
+    }
+    archiveArtifacts allowEmptyArchive: true, artifacts: 'kola-testiso*.tar.xz'
+
+    stage("Test Live Images") {
+        parallel(testIsoRuns)
+    }
+}


jlebon:
Let's make this its own custom step instead? E.g. `fcosKolaTestIso.groovy`?

(Eventually we should rename `fcosBuild()` to e.g. `cosaBuild()`, and `fcosKola()` to `kolaRun()` if we're going to use the library to build RHCOS too. Then this one would be `kolaTestIso()`.)

ravanelli:
@jlebon Thanks for the review again ;). 
Done, changed it to fcosKolaTestIso

vars/fcosKolaTestIso.groovy
```diff
@@ -0,0 +1,28 @@
...
+
+        try {
+            parallel metal: {
+                shwrap("cd ${cosaDir}  && env TMPDIR=\$(pwd)/tmp/ kola testiso -S ${extraArgs} --output-dir tmp/kola-testiso-metal")
+            }
+            if (!params['skipMetal4k']) {
+                metal4k: {
+                    shwrap("cd ${cosaDir} && env TMPDIR=\$(pwd)/tmp/ kola testiso -S ${extraArgs} --output-dir tmp/kola-testiso-metal4k")
+                }
+            }
```
jlebon, line 19: Hmm, I'm not sure if this works. I think you have to make a map first and conditionally add `metal4k` in that map, and then call `parallel` on the map. See in `fcosKola`.

vars/fcosKolaTestIso.groovy
```diff
@@ -0,0 +1,28 @@
...
+//     extraArgs:      string  -- extra arguments to pass to `kola testiso`
+//     skipMetal4k:    boolean -- skip metal4k image
+def call(params = [:]) {
+    stage("Test Live Images") {
+        def cosaDir = utils.getCosaDir(params)
+        def extraArgs = params.get('extraArgs', "");
+
+        try {
+            parallel metal: {
+                shwrap("cd ${cosaDir}  && env TMPDIR=\$(pwd)/tmp/ kola testiso -S ${extraArgs} --output-dir tmp/kola-testiso-metal")
```
jlebon, line 13: Why do we need the `TMPDIR` again?

vars/fcosKolaTestIso.groovy
ravanelli: I have tested it in both cases, using and not using the skipMetal4k.
Here is the output for it in Jenkins.

`[Pipeline] parallel
[Pipeline] { (Branch: metal)
[Pipeline] withEnv
[Pipeline] {
[Pipeline] sh
+ set -xeuo pipefail
++ umask
+ '[' 0022 = 0000 ']'
+ cd /srv
++ pwd
+ env TMPDIR=/srv/tmp/ kola testiso -S --output-dir tmp/kola-testiso-metal
Testing scenarios: [pxe-install iso-offline-install pxe-offline-install]
Successfully tested scenario pxe-install for 32.20201104.dev.0 on bios (metal)
Successfully tested scenario pxe-offline-install for 32.20201104.dev.0 on bios (metal)
Successfully tested scenario iso-offline-install for 32.20201104.dev.0 on bios (metal)
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // parallel
[Pipeline] withEnv
[Pipeline] {
[Pipeline] sh
+ set -xeuo pipefail
++ umask
+ '[' 0022 = 0000 ']'
+ cd /srv
+ kola testiso -S --qemu-native-4k --no-pxe --output-dir tmp/kola-testiso-metal4k
Testing scenarios: [iso-offline-install]
Successfully tested scenario iso-offline-install for 32.20201104.dev.0 on uefi (metal4k)
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] withEnv`

Anyway, I can make it using map as we have in `fcosKola`

vars/fcosKolaTestIso.groovy
ravanelli: Meh, looks in my tests we don't really need it at all indeed, checked the other CIs and they are not using it. I was trying to keep it as similar as possible from what we have in cosa ci for now.

vars/fcosKolaTestIso.groovy
jlebon: Hmm interesting that Jenkins doesn't error out on this. But anyway, I think what's it's actually doing there is just running the `metal` branch on its own and then running the `metal4k` code separately. Normally it would've printed something like this (compare with the output of an `fcosKola` run):

```
[Pipeline] parallel
[Pipeline] { (Branch: metal)
[Pipeline] { (Branch: metal4k)
```

`parallel` takes a map and here we're passing it a map with a single key (`metal`).

vars/fcosKolaTestIso.groovy
```diff
@@ -0,0 +1,34 @@
...
+        } finally {
+            shwrap("cd ${cosaDir} && tar -cf - tmp/kola-testiso-metal/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-metal.tar.xz")
+        }
+    }
+    if (!params['skipMetal4k']) {
+        testIsoRuns["Metal4k"] = {
+            try {
+                shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k ${extraArgs4k} --output-dir tmp/kola-testiso-metal4k")
+            } finally {
+                shwrap("cd ${cosaDir} && tar -cf - tmp/kola-testiso-metal4k/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-metal4k.tar.xz")
```
jlebon, line 25: I think we need `--no-pxe` here since `kola testiso` doesn't support testing metal4k + PXE yet.

vars/fcosKolaTestIso.groovy
```diff
@@ -0,0 +1,34 @@
...
+//     extraArgs:      string  -- extra arguments to pass to `kola testiso`
+//     extraArgs4k:    string  -- extra arguments to pass to 4k `kola testiso`
+//     skipMetal4k:    boolean -- skip metal4k image
+def call(params = [:]) {
+    def cosaDir = utils.getCosaDir(params)
+    def extraArgs = params.get('extraArgs', "");
+    def extraArgs4k = params.get('extraArgs4k', "");
+
+    testIsoRuns = [:]
+    testIsoRuns["Metal"] = {
```
jlebon, line 13: Super minor: let's call the branches `metal` and `metal4k`? That matches the lowercase of `fcosKola`'s `run` and `run-upgrade`.

jlebon (approved):
A few minor nits, but LGTM overall!

vars/fcosKolaTestIso.groovy
ravanelli: I'm thinking to pass it as argument.
In RHCOs for example we have:
`kola testiso -S --qemu-native-4k --scenarios iso-install --output-dir tmp/kola-metal4k`
It will be more flexible, what do you think?

jlebon: Sure SGTM. Was thinking we could have a `scenarios` variable to make this nicer, but I think it's good anyway to have an `extraArgs` escape hatch. So we can add a `scenarios` var as well as a follow-up. (Or if you feel like it, feel free!)

jlebon:
Your PR here drove me to do https://github.com/coreos/coreos-ci-lib/pull/49.

ravanelli:
> Your PR here drove me to do #49.

@jlebon In this case we will have the Kola stage + run + run_upgrade stages. I was thinking in actually keep the branches and remove the stages run + run_updgrades, in this case we would have Kola stage and the rest inside it, since it is part of the Kola tests. For cosa and rhcos we don't have run + run_updgrade as stages, just Kola (QEMU). But I'm fine with that ;)

vars/fcosKolaTestIso.groovy
```diff
@@ -0,0 +1,46 @@
...
+    def scenarios4k = params.get('scenarios4k', "");
+
+    testIsoRuns = [:]
+    testIsoRuns["metal"] = {
+        try {
+            if (params['scenarios']) {
+                shwrap("cd ${cosaDir}  && kola testiso -S ${extraArgs} --scenarios ${scenarios} --output-dir tmp/kola-testiso-metal")
+            } else {
+                shwrap("cd ${cosaDir}  && kola testiso -S ${extraArgs} --output-dir tmp/kola-testiso-metal")
+            }
```
jlebon, line 23: I think we could simplify this by just have `--scenarios` be part of the `scenarios` variable if it's non-zero, but we can do that as a follow-up!

jlebon (approved):
Thanks a lot!

PR #49, opened 2020-11-06
fcosKola: don't nest branches in "stage"

jlebon:
These stage labels aren't actually getting rendered. Instead, the branch
name is used, which is good enough. So let's simplify this.


jlebon:
Best viewed with whitespace change ignored.

jlebon:
@ravanelli noticed that I was removing the branch instead of the stage in the upgrade test section (thanks!). Fixed that now.

ravanelli (approved):
LGTM

PR #50, opened 2020-11-06
utils: Add cosaCmd

ravanelli:
Add cosa_cmd in order to easily run cosa commands

### DIFF
--- a/vars/utils.groovy
+++ b/vars/utils.groovy
@@ -1,3 +1,18 @@
+def cosaCmd(params = [:]) {
+// Available parameters:
+//     cosaDir:        string  -- cosa working directory
+//     args:           string  -- arguments to pass to `cosa`
+//     user:           string  -- user to run the cosa commands 
+    def cosaDir = getCosaDir(params)
+    def args = params.get('args', "");
+
+    if(!params['user']) {
+        shwrap("cd ${cosaDir} && cosa ${args}")
+    } else {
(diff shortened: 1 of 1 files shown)


vars/utils.groovy
```diff
@@ -1,3 +1,11 @@
+def cosa_cmd(cosaDir, userBuilder, args) {
```
jlebon, line 1: How about `cosaCmd` to match the casing of other commands?

Also, maybe better to make this also use the `params = [:]` paradigm, that way users can omit `cosaDir` and `userBuilder` if the defaults are fine (i.e. `/srv/fcos` and no `sudo -u $user`). WDYT?

vars/utils.groovy
```diff
@@ -1,3 +1,14 @@
+def cosaCmd(params =[:]) {
```
jlebon, line 1:
```suggestion
def cosaCmd(params = [:]) {
```

vars/utils.groovy
```diff
@@ -1,3 +1,14 @@
...
+    def cosaDir = getCosaDir(params)
+    def args = params.get('args', "");
+
+    if(!params['userBuilder']) {
```
jlebon, line 5: Let's just call this parameter `user`?

vars/utils.groovy
jlebon, line 1: Maybe a docstring before this line listing the supported parameters?

jlebon (approved):

PR #51, opened 2020-11-11
Add fcosCosaBuildMetal

ravanelli:
Make fcosCosaBuildMetal available as a custom step

### DIFF
--- a/vars/fcosCosaBuildMetal.groovy
+++ b/vars/fcosCosaBuildMetal.groovy
@@ -1,0 +1,25 @@
+// Available parameters:
+// Run build metal on cosa
+//    cosaDir:        string  -- cosa working directory
+//    skipMetal4k:    boolean -- skip metal4k image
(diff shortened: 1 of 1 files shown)


jlebon:
Re. this PR and #53 and #54, I'm not sure how useful these custom steps are. They look like they're abstracting away very little logic. Here are two ideas on how to further streamline these cosa commands for callers:
1. let's make `cosaCmd` an actual custom step. That way we can just do `cosaCmd(...)` instead of `utils.cosaCmd(...)`.
2. let's generalize the idea in #52 to a custom step like `cosaParallelCmds(...)` so that you can run any `cosa` commands you want in parallel.

WDYT?

ravanelli:
@jlebon Yes, I think it make much more sense! Thanks for the reviews again. I'm closing this PR since I went to the ideia in #52  to create a custom step cosaParallelCmds.

PR #52, opened 2020-11-12
Add cosaParallelCmds.groovy

ravanelli:
Make fcosCosaBuildMetal available as a custom step

### DIFF
--- a/vars/cosaParallelCmds.groovy
+++ b/vars/cosaParallelCmds.groovy
@@ -1,0 +1,13 @@
+// Run cosa commands in parallel
+// Available parameters:
+//    cosaDir:        string    -- cosa working directory
+//    commands:     []string    -- list of commands to run
+//    user:           string    -- user account to run cosa commands
+
+def call(params = [:]) {
+    def commands = params['commands'];
+
+    parallel commands.inject([:]) { d, i -> d[i] = {
+        utils.cosaCmd(cosaDir: params['cosaDir'], user: params['user'], args: "buildextend-${i.toLowerCase()}")
+    }; d }
+}


vars/cosaParallelCmds.groovy
```diff
@@ -0,0 +1,27 @@
+// Run cosa commands in parallel
+// Available parameters:
+//    cosaDir:        string    -- cosa working directory
+//    isBuild:        boolean   -- run build commands using `cosa buildextend-`
+//                                 e.g pass to runCommands only:
+//                                 ['metal', 'metal4k', 'live', 'AWS']
```
jlebon, line 6: Do you expect to use this for commands which aren't `buildextend-...`? I think any other command doesn't parallelize as obviously.

vars/cosaParallelCmds.groovy
```diff
@@ -0,0 +1,27 @@
...
+//    runCommands:  []string    -- list of commands to run
```
jlebon, line 7: Maybe just `commands`?

vars/cosaParallelCmds.groovy
```diff
@@ -0,0 +1,27 @@
...
+//    user:           string    -- user account to run cosa commands
+
+def call(params = [:]) {
+    def runCommands = params['runCommands'];
+
+    if (params['stage']) {
+        stage(params['stage'])
+    } else {
+        stage("Run")
+    }
```
jlebon, line 18: Does this actually wrap the parallel stages or does it just result in an empty stage before the parallel runs? Normally, to define a stage you'd pass the stage code as a close (i.e. `stage("foo") { ... }`).

Maybe let's rework this as:

```groovy
stageName = params.get('stage', "Run")
stage(stageName) {
    parallel ...
}
```

?

jlebon: Actually, I think we should just delete that ability entirely. Let's leave it up to the caller to wrap the call to `cosaParallelCmds` in a stage if they'd like. WDYT?

jlebon (approved):

PR #55, opened 2020-11-12
fcosKola: add `addExtTests` parameter

ravanelli:
Add self tests in fcosKola

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -1,5 +1,6 @@
 // Run kola tests on the latest build in the cosa dir
 // Available parameters:
+//    addExtTests:    []string -- list of test paths to run
 //    cosaDir:         string  -- cosa working directory
 //    parallel:        integer -- number of tests to run in parallel (default: 8)
 //    skipUpgrade:     boolean -- skip running `cosa kola --upgrades`
@@ -34,6 +35,12 @@ def call(params = [:]) {
         }
         def parallel = params.get('parallel', 8);
         def extraArgs = params.get('extraArgs', "");
+        def addExtTests = params.get('addExtTests', [])
+
+        for (path in addExtTests) {
+            args += " --exttest ${env.WORKSPACE}/${path}"
+        }
+
         try {
             if (params['basicScenarios']) {
                 shwrap("cd ${cosaDir} && cosa kola run --basic-qemu-scenarios")


vars/fcosKola.groovy
```diff
@@ -3,6 +3,7 @@
 //    cosaDir:         string  -- cosa working directory
 //    parallel:        integer -- number of tests to run in parallel (default: 8)
 //    skipUpgrade:     boolean -- skip running `cosa kola --upgrades`
+//    skipSelfTests:   boolean -- skip running `run-kola-self-tests`
```
jlebon, line 6: How about we flip this around and have the self-test be disabled by default? So we'd call this e.g. `selfTest`.

jlebon: My reason for this is that I think the self-test isn't really useful for the majority of upstream projects.

ravanelli: @jlebon Thanks again for the reviews!
Done ;)

vars/fcosKola.groovy
```diff
@@ -58,6 +59,16 @@ def call(params = [:]) {
             shwrap("test -d ${cosaDir}/tmp/kola-upgrade")
         }
     }
+    if (params["selfTest"]) {
+        kolaRuns['self_tests'] = {
+            try {
+                shwrap("cd ${cosaDir} && ${env.WORKSPACE}/ci/run-kola-self-tests")
```
jlebon, line 65: Sorry, I missed this before. I don't think we should rely on this script here. If we want to support the self-test as target in `fcoSKola`, we shouldn't assume that it's going to be running against a cosa git checkout. (E.g. another place where I see this being useful is in the https://github.com/coreos/fedora-coreos-config CI since it has the code that gathers the journal logs).

So here I think we should call out to `kola` ourselves like that script does.

vars/fcosKola.groovy
```diff
@@ -3,6 +3,7 @@
 //    cosaDir:         string  -- cosa working directory
 //    parallel:        integer -- number of tests to run in parallel (default: 8)
 //    skipUpgrade:     boolean -- skip running `cosa kola --upgrades`
+//    selfTests:      []string -- list of test paths to run
```
jlebon, line 6: Ahh OK, one thing I didn't realize here was that the self-tests themselves were located in cosa under `tests/kola-ci-self` and not as part of kola itself.

Hmm, in that case, how about we generalize this slightly so that it's just a parameter that takes paths that we pass to the main kola run. E.g. something like

```groovy
// addExtTests []string -- list of dirs containing additional external tests
...
def call(params = [:]) {
    ...
    def addExtTests = params,get('addExtTests', [])
    ...
    for (path in addExtTests) {
        args += "--exttest ${path}"
    }
    ...
}
```

?

Then, CI for cosa can just pass the `tests/kola-ci-self` dir when it runs kola, or if it really wants the run to be separate from the main run, call `fcosKola` again with that and an `extraArgs` set to `ext.kola-ci-self*` so it only runs those tests.

Does that make sense?

vars/fcosKola.groovy
```diff
@@ -34,6 +35,12 @@ def call(params = [:]) {
         }
         def parallel = params.get('parallel', 8);
         def extraArgs = params.get('extraArgs', "");
+        def addExtTests = params.get('addExtTests', [])
+
+        for (path in addExtTests) {
+            args += "--exttest ${env.WORKSPACE}/${path}"
```
jlebon, line 41: I think we need a leading space here to make sure that they're separate args. I.e.

```suggestion
            args += " --exttest ${env.WORKSPACE}/${path}"
```

jlebon (approved):
LGTM!

PR #58, opened 2020-12-07
Change default cosa --strict parameter

ravanelli:
In order to be more flexible across the CIs
remove the `--strict` default option. And let
who is calling it define the parameters.
We have cases where we don't want to use
`--strict`.

### DIFF
--- a/vars/fcosBuild.groovy
+++ b/vars/fcosBuild.groovy
@@ -9,9 +9,13 @@
 //    extraArgs:      string  -- extra arguments to pass to `cosa build`
 //    cosaDir:        string  -- cosa working directory
 //    noForce:        boolean -- don't force a cosa build even if nothing changed
+//    noStrict        boolean -- do not run cosa using `--strict' option
 def call(params = [:]) {
     stage("Build FCOS") {
         def cosaDir = utils.getCosaDir(params)
+        def extraFetchArgs = params.get('extraFetchArgs', "");
+        def extraArgs = params.get('extraArgs', "");
+
         shwrap("mkdir -p ${cosaDir}")
 
         if (!params['skipInit']) {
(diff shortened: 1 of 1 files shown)


vars/fcosBuild.groovy
```diff
@@ -33,14 +33,14 @@ def call(params = [:]) {
             }
         }
 
-        def extraFetchArgs = params.get('extraFetchArgs', "");
-        shwrap("cd ${cosaDir} && cosa fetch --strict ${extraFetchArgs}")
+        def extraFetchArgs = params.get('extraFetchArgs', "--strict");
```
jlebon, line 36: I think normally you do always want strict mode turned on in CI because it ensures that the lockfiles are the complete sources of truth (see https://github.com/coreos/rpm-ostree/pull/1858 and https://github.com/coreos/fedora-coreos-tracker/issues/454). Essentially, non-strict mode is useful mostly for the developer case.

In the case of RHCOS, which doesn't use lockfiles yet, using `cosa {fetch,build} --strict` should have no effect. 

That said, it does make sense to make this configurable to be flexible (and e.g. allow developers to hack up the manifest quickly and fire up a Jenkins job for it).

Instead of putting this in `extraFetchArgs` and `extraArgs`, and potentially having callers which already use these lose the `--strict` arg, WDYT about adding a new `strict` bool instead which defaults to `true`?

ravanelli: Thanks for the overview @jlebon ;)

jlebon (approved):
Minor bikeshed, otherwise LGTM! :)

vars/fcosBuild.groovy
```diff
@@ -2,6 +2,7 @@
 // Available parameters:
 //    make:           boolean -- run `make && make install DESTDIR=...`
 //    skipInit:       boolean -- assume `cosa init` has already been run
+//    skipStrict      boolean -- do not run cosa using `--strict' option
```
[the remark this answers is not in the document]
ravanelli: aaa good point. Fixed!

jlebon (approved):

PR #59, opened 2020-12-29
fcosBuild: Change to use cosaCmd

ravanelli:


vars/fcosBuild.groovy
```diff
@@ -19,7 +20,8 @@ def call(params = [:]) {
         shwrap("mkdir -p ${cosaDir}")
 
         if (!params['skipInit']) {
-            shwrap("cd ${cosaDir} && cosa init https://github.com/coreos/fedora-coreos-config")
+            params['args'] = "init https://github.com/coreos/fedora-coreos-config"
+            utils.cosaCmd(params)
```
darkmuggle, line 24: I think it would be cleaner to separate the `params` from config/state. For readability, having the second param of `args` would be cleaner and improve the readability.  Or go-full OOO and have `cosaCmd` be an object that is created with the parameters. In this case, OOO might be worth the trouble since we're wrapping COSA. 

```
class cosaCmd {
   String CosaDir
   String User

   def run(args) {
         if(!params['user']) {
            return shwrap("cd ${CosaDir} && cosa ${args}")
         } else {
            return shwrap("cd ${CosaDir} && sudo -u ${User} cosa ${args}")
        }
    }

}

```

And then you would use:
```
def cosaCommander = new utils.cosaCmd("/srv", "builder")
cosaCommander.run("init ....")
```

jlebon:
This is for https://github.com/coreos/coreos-assembler/pull/1931, right?

So the base issue here is that we want cosa's CI to test the unprivileged path, but we currently provision the pod as privileged so that we can build and install cosa. As mentioned in https://github.com/coreos/coreos-assembler/pull/1931/files#r540481407, I think ideally we'd be able to separate out those parts of CI better.

/some time passes

OK, I think I've figured out a way to get something like `asUser() { ... }` as discussed there:
1. we enhance the script in `shwrap()` to just check an env var (let's say `AS_USER`) and if defined, re-run itself as that user (i.e. `exec sudo -u "${AS_USER}" "\$0" "\$@"`)
2. we add a new `asUser(user, Closure body)` custom step which just wraps the closure passed with `withEnv(["AS_USER=${user}])`.

Then, cosa's CI can just do the first bits where it downloads packages and installs cosa privileged, and everything else under an `asUser("builder") { ... }` block. WDYT?

ravanelli:
@jlebon Actually for  coreos/coreos-assembler#1931 I'm testing a way from building the container and then using it in the others steps. I checked the imagestream doc, and basically it will work as cosa already works, the build  (the builds steps) will result in an image as output. For this patch I'm trying to reflect new functions in utils.

jlebon:
> @jlebon Actually for [coreos/coreos-assembler#1931](https://github.com/coreos/coreos-assembler/pull/1931) I'm testing a way from building the container and then using it in the others steps. I checked the imagestream doc, and basically it will work as cosa already works, the build (the builds steps) will result in an image as output.

Heh yeah, I was looking at that too. Just wasn't sure if it'd require widening ACLs. My other concern there is making sure we figure out GC so that the imagestream isn't just littered with PR tests (this is where the Prow model really shines, having a separate namespace for each PR). Otherwise yeah, if we can get it to work without too much trouble, I think this is the cleanest approach.

In that case though, we wouldn't need this PR at all because then we'd provision an unprivileged pod anyway and everything would run as that unprivileged user.

ravanelli:
Closing due https://github.com/coreos/coreos-ci-lib/pull/61

PR #61, opened 2021-01-15
Add buildImage

ravanelli:
Create a buildconfig and trigger the build for
coreos-assembler creating a imageStream for it.

It allows us to split the build process in 2 phases.
The coreos-assembler build will be done via DockerFile
the same way we have in most of our CIs today.

### DIFF
--- a/resources/com/github/coreos/buildconfig.json
+++ b/resources/com/github/coreos/buildconfig.json
@@ -1,0 +1,101 @@
+{
+  "kind":"Template",
+  "apiVersion":"v1",
+  "metadata":{
+    "name":"${NAME}"
+  },
+  "labels":{
+    "template":"${NAME}",
+    "app":"${NAME}"
+  },
+  "objects":[
+    {
+      "kind":"ImageStream",
+      "apiVersion":"v1",
+      "metadata":{
+        "name":"${NAME}",
+        "ownerReferences":[
+          {
+            "apiVersion":"batch/v1",
+            "kind":"Job",
+            "name":"${OWNERNAME}",
+            "uid":"${UID}",
+            "controller":true
+          }
+        ]
+      },
+      "spec":{
+        "failedBuildHistoryLimit":1,
+        "successfulBuildsHistoryLimit":1,
+        "lookupPolicy":{
+          "local":true
+        }
+      }
+    },
+    {
+      "kind":"BuildConfig",
+      "apiVersion":"v1",
+      "metadata":{
+        "name":"${NAME}",
+        "ownerReferences":[
+          {
+            "apiVersion":"batch/v1",
+            "kind":"Job",
+            "name":"${OWNERNAME}",
+            "uid":"${UID}",
+            "controller":true
+          }
+        ]
+      },
+      "spec":{
+        "source":{
+          "type":"Git",
+          "git":{
+            "uri":"none",
+            "ref":"none"
+          }
+	},
+        "strategy":{
+          "dockerStrategy":{
+            "dockerfilePath":"${DOCKERFILE}"
+          },
+          "dockerfilePath":"${DOCKERFILE}"
+        },
+        "output":{
+          "to":{
+            "kind":"ImageStreamTag",
+            "name":"${NAME}:latest",
+            "namespace":"coreos-ci"
+          }
+        }
+      }
+    }
+  ],
+  "parameters":[
+    {
+      "name":"NAME",
+      "displayName":"Name",
+      "description":"The name assigned to all of the frontend objects defined in this template.",
+      "required":true,
+      "value":"none"
+    },
+    {
+      "name":"DOCKERFILE",
+      "displayName":"Docker File",
+      "description":"The Docker file that will be use to build the application",
+      "value":"Dockerfile"
+    },
+    {
+      "name":"OWNERNAME",
+      "displayName":"Owner Name",
+      "description":"Owner Name for cascading delition",
+      "value":"none"
+    },
+    {
+      "name":"UID",
+      "displayName":"Owner uid",
+      "description":"Onwer uid for cascading delition",
+      "value":"none"
+    }
+  ]
+}
--- a/resources/com/github/coreos/job.yaml
+++ b/resources/com/github/coreos/job.yaml
@@ -1,0 +1,24 @@
+kind: Template
+apiVersion: v1
+metadata:
+  name: clean-up
+objects:
+- apiVersion: batch/v1
+  kind: Job
+  metadata:
+    name: "${NAME}"
+  spec:
+    ttlSecondsAfterFinished: 100
+    template:
+      spec:
+        containers:
+        - name: "${NAME}"
+          image: fedora:33
+          command:
+          - sleep
+          - 2h
+        restartPolicy: Never
+parameters:
+- name: NAME
+  required: true
+  value: cleanupJob
--- a/vars/buildImage.groovy
+++ b/vars/buildImage.groovy
@@ -1,0 +1,70 @@
+// Builds a container image and returns its name.
+// Available parameters:
+//    dockerFile    string -- DockerFile used for the buildconfig
+//    workspace     string -- Path for local source dir used for `--from-dir=`
+def call(params = [:]) {
+    def imageName
+    def bcObj
+
+    openshift.withCluster() {
+        openshift.withProject() {
+            stage('CleanUp') {
+            // Once we have the TTL controller enabled we can remove this stage
+            // More: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/
+                def jobs = openshift.selector("jobs").names()
+                for (job in jobs) {
+                    if (job.startsWith('job/coreos-ci')) {
+                        def obj = openshift.selector(job).object()
+                        if (obj.status.completionTime) {
+                            openshift.selector(job).delete()
+                        }
+                    }
+                }
+            }
+            stage('Prepare Env') {
+                // TODO: Change buidlconfig to yaml and to use openshift.process
+                // Issue described in:
+                // https://github.com/coreos/coreos-ci-lib/pull/61#discussion_r569401940
+                def bcJSON  = libraryResource 'com/github/coreos/buildconfig.json'
+                def jobYaml = libraryResource 'com/github/coreos/job.yaml'
+                def jobObj  = readYaml text: jobYaml
+                bcObj       = readJSON text: bcJSON
+
+                def UUID    = UUID.randomUUID()
+                def scmVars = checkout scm
+                def repo    = (scmVars.GIT_URL.substring(scmVars.GIT_URL.lastIndexOf('/') + 1)).replace('.git','')
+
+                jobObj      = openshift.process(jobObj, "-p", "NAME=coreos-ci-$repo-${UUID}")
+                def app     = openshift.create(jobObj)
+                def job     = app.narrow('job')
+                def jobMap  = job.object()
+                def uid     = jobMap.metadata.get('uid')
+
+                imageName = "coreos-ci-$repo-${UUID}"
+
+                //Create and unique name for image and bc
+                bcObj['parameters'][0] +=['name': 'NAME', 'value': "coreos-ci-$repo-${UUID}".toString()]
+
+                //Add OwnerReference for cascading deletion with the Job timeout
+                bcObj['parameters'][3] +=['name': 'UID', 'value': "${uid}".toString()]
+                bcObj['parameters'][2] +=['name': 'OWNERNAME', 'value': "coreos-ci-$repo-${UUID}".toString()]
+                if (params['dockerFile']) {
+                    bcObj['parameters'][1] +=['name': 'DOCKERFILE', 'value': "${params['dockerFile']}".toString()]
+                }
+            }
+            stage('Create BuildConfig') {
+                openshift.create(openshift.process(bcObj))
+            }
+            stage('Build') {
+                def workspace = params.get('workspace', ".");
+                shwrap("tar --exclude=./pod* -zcf /tmp/dir.tar ${workspace}")
+                def bc = openshift.selector('bc', imageName).startBuild("--from-archive=/tmp/dir.tar")
+
+                // Showing logs in Jenkins is also a way
+                // to wait for the build to finsih
+                bc.logs('-f')
+           }
+        }
+    }
+    return imageName.toString()
+}


jlebon:
First, thanks a lot for writing and testing this! So far CoreOS CI hasn't really been "OpenShift-aware" and it makes a lot of sense to have support for building images and then running tests in them (it basically takes it much closer to the Prow model).

I have a few concerns about this. Notably:
- AFAICT, there's nothing actually linking the source ref to the PR under test, right? Won't this always just build from git master of coreos-assembler? I would expect `buildImage()` to pass the working directory or something to `oc build` (not sure if `startBuild()` has sugar for this).
- How does permission setting look like? I'm guessing this requires enhancing RBAC for the `jenkins` SA to allow it to create ImageStreams and BuildConfigs and also start builds?
- Right now all builds are outputting to `cosa-image:latest`. But this means that concurrent PRs will step on each other. I think what we need here is to have PRs output to separate tags of the image stream. But I'm not sure if it's possible to override the output tag at `oc start-build` time. Or maybe we just create separate BuildConfigs for each PR, each outputting to a different tag.
- In light of the previous item, we need to think about garbage collection. With Prow for example, it allocates a full namespace and it just tears it down at the end. Here I think we'd need some sort of janitor job in CoreOS CI or something which prunes old BuildConfigs and ImageStream tags?

Also, similarly to https://github.com/coreos/coreos-ci-lib/pull/45, I think there's a mismatch of abstraction here. A `buildImage` type functionality IMO should be generic over all types of images instead of being specifically about building cosa. So e.g. it could take a git URL and ref and a Dockerfile path that it uses to fill in a BuildConfig manifest before instantiating it. This is similar to how it creates pods from `pod.json`.

ravanelli:
> * AFAICT, there's nothing actually linking the source ref to the PR under test, right? Won't this always just build from git master of coreos-assembler? I would expect `buildImage()` to pass the working directory or something to `oc build` (not sure if `startBuild()` has sugar for this).

Ooo, yes totally forgot to add this, my fault. I already tested it lol , since I'm able to get the branch that called it using env.GIT_BRANCH I can replace the gitbranch in the json file. Let me get that added here.

> * How does permission setting look like? I'm guessing this requires enhancing RBAC for the `jenkins` SA to allow it to create ImageStreams and BuildConfigs and also start builds?

Yes, this part definitely needs more test. I don't have access in the ocp or neither in Jenkins where the cosa CI is running. However, in the CentOs cluster where I tested it I do have access to create bc, imageStream and such. I would need to run it through the cosa CI to validate it first.

> * Right now all builds are outputting to `cosa-image:latest`. But this means that concurrent PRs will step on each other. I think what we need here is to have PRs output to separate tags of the image stream. But I'm not sure if it's possible to override the output tag at `oc start-build` time. Or maybe we just create separate BuildConfigs for each PR, each outputting to a different tag.

Yes, it is possible. I think I can easily replace the tag in the json file using the cosa branch name that called it.


> * In light of the previous item, we need to think about garbage collection. With Prow for example, it allocates a full namespace and it just tears it down at the end. Here I think we'd need some sort of janitor job in CoreOS CI or something which prunes old BuildConfigs and ImageStream tags?

We can delete the imageStream and the BuildConfig at the end of the job, not sure if it would be enough.. if the job fails for example. Let me check the docs and see how hard would be to have a namespace for each PR.

> Also, similarly to #45, I think there's a mismatch of abstraction here. A `buildImage` type functionality IMO should be generic over all types of images instead of being specifically about building cosa. So e.g. it could take a git URL and ref and a Dockerfile path that it uses to fill in a BuildConfig manifest before instantiating it. This is similar to how it creates pods from `pod.json`.

I added the jsonfile path as parameter, the json file would point to the git URL, branch and the DockerFile name, in this case you can trigger whatever you want using your buildconfig.

ravanelli:
>Also, similarly to #45, I think there's a mismatch of abstraction here. A buildImage type functionality IMO should be generic over all types of images instead of being specifically about building cosa. So e.g. it could take a git URL and ref and a Dockerfile path that it uses to fill in a BuildConfig manifest before instantiating it. This is similar to how it creates pods from pod.json.

Reading it again.  Got it better, yeah I can make it in the way the pod was done.

jlebon:
> We can delete the imageStream and the BuildConfig at the end of the job, not sure if it would be enough.. if the job fails for example.

Right yeah, because of the failure case we would need to have some code to make sure things eventually get cleaned up.

Hmm, here's an idea: we create a sleeper pod which just does e.g. `sleep $((24 * 3600))` (i.e. sleeps for a day before exiting). Then when we create the ImageStream and BuildConfig objects, we mark them as owned by the sleeper pod via [`ownerReferences`](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/). This means that when the sleeper pod is deleted, the ImageStream and BuildConfig will automatically be deleted.

Then, all we would do is whenever `buildImage()` is called, it deletes any sleeper pods which have already exited. Ideally, k8s would have an "auto-delete" pod feature, though doesn't seem like that exists yet (ahh, it looks like `Job`s do have alpha support for that: https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically, so that might be even cleaner!).

ravanelli:
@jlebon btw I wasn't really able to test the `TTL controller`, since it is an Alpha feature it needs to be enabled. I need to dig more to understand if I can enabled it for testing in the cluster I am. However the yaml files were created ok and if you delete the job the cascading delete will happen.

vars/buildImage.groovy
```diff
@@ -0,0 +1,99 @@
...
+    pipeline {
+        // We don't need any specify agent just to add the buildconfig and trigger it,
+        // this pod will be destroyed in the end
+        agent {
+            node { label 'nodejs' }
+        }
+        stages {
+            stage('Prepare Env') {
+                steps{
+                   script{
```
jlebon, line 18: Can we use scripted pipeline instead? We shouldn't have to provision an agent either; let's leave that to callers. So e.g. this would be something like:

```groovy
def call(params = [:]) {
    openshift.withCluster() {
        openshift.withProject() {
            // load JSONs and set their values
            // create objects
            // follow logs
        }
    }
```

I left out the stages. I think it'probably fine without them, but you can include them as stages nested within the `openshift.withProject()` closure if you'd like.

resources/com/github/coreos/buildconfig.json
```diff
@@ -0,0 +1,163 @@
...
+      "description": "The name assigned to all of the frontend objects defined in this template.",
+      "required": true,
+      "value": "cosa-image"
+    },
+    {
+      "name": "NAMESPACE",
+      "displayName": "Namespace",
+      "description": "The OpenShift Namespace where the ImageStream resides.",
+      "required": true,
+      "value": "fedora-coreos"
```
jlebon, line 117: This should default to `coreos-ci` instead, since that's the name of the project where all the upstream CI runs.

resources/com/github/coreos/buildconfig.json
```diff
@@ -0,0 +1,163 @@
...
+    },
+    {
+      "name": "SOURCE_REPOSITORY_URL",
+      "displayName": "Git Repository URL",
+      "description": "The URL of the repository with your application source code.",
+      "required": true,
+      "value": "https://github.com/coreos/coreos-assembler.git"
```
jlebon, line 124: Can we not provide a default at all for this? Let's make it a hard requirement that callers have to provide it.

resources/com/github/coreos/job.json
```diff
@@ -0,0 +1,52 @@
+{
```
jlebon, line 1: Couldn't we fold this into the `buildconfig.json` template?

resources/com/github/coreos/job.json
```diff
@@ -0,0 +1,52 @@
...
+  },
+  "objects": [
+  {      
+  "apiVersion": "batch/v1",
+  "kind": "Job",
+  "metadata": {
+    "name": "${NAME}"
+  },
+  "spec": {
+    "ttlSecondsAfterFinished": 10800,
```
jlebon, line 20: If this isn't fully supported yet, let's go with the other option for now? (A sleeper pod, and then some code to delete exited sleeper pods.)

vars/buildImage.groovy
```diff
@@ -0,0 +1,99 @@
...
+                       openshift.withCluster() {
+                           openshift.withProject() {
+
+                               def bcJSON  = libraryResource 'com/github/coreos/buildconfig.json'
+                               def jobJSON = libraryResource 'com/github/coreos/job.json'
+                               def jobObj  = readJSON text: jobJSON
+                               def bcObj   = readJSON text: bcJSON
+                               def tagName = UUID.randomUUID()
+
+                               jobObj['parameters'][0] +=['name': 'NAME', 'value': "cleanup-${tagName}".toString()]
```
jlebon, line 28: Instead of setting parameter values like this, we can do like described here: https://github.com/openshift/jenkins-client-plugin#cannot-live-without-openshift-templates-no-problem and use `oc.process(obj, "-p", "NAME=...")`.

vars/buildImage.groovy
```diff
@@ -0,0 +1,99 @@
...
+
+                               writeYaml(file: "job-${tagName}.yaml", data: jobObj, overwrite: true)
+                               jobTemplatePath = env.WORKSPACE+"/job-${tagName}.yaml"
```
jlebon, line 31: If I'm reading https://github.com/openshift/jenkins-client-plugin#cannot-live-without-openshift-templates-no-problem right, I think we can avoid writing to disk entirely and just instantiate the template via `openshift.create(openshift.process(...))`.

vars/buildImage.groovy
```diff
@@ -0,0 +1,99 @@
...
+                               bcObj['parameters'][8] +=['name': 'UID', 'value': "${uid}".toString()]
+                               bcObj['parameters'][7] +=['name': 'OWNERNAME', 'value': "cleanup-${tagName}".toString()]
+
+                               if (params['namespace']) {
+                                   bcObj['parameters'][1] +=['name': 'NAMESPACE', 'value': "${params['nameSpace']}"]
+                               }
+                               if (params['gitURL']) {
+                                   bcObj['parameters'][2] +=['name': 'SOURCE_REPOSITORY_URL', 'value': "${params['gitURL']}"]
+                               }
+                               if (env.BRANCH_NAME && env.BRANCH_NAME.startsWith('PR-')) {
```
jlebon, line 52: Oh, does this work? I've had trouble in the past accessing environment variables from Jenkins shared libraries.

jlebon (commented):
This is very cool stuff!

resources/com/github/coreos/job.json
ravanelli: To be honestly I tried. But I couldn't find an easier way to do it. Probably due the lack of experiencie with ocp. ='( I think we would need to find a way for the Job to create the bc, and the imageStream or to trigger the build, so it would get the ownerReference. Probably we would need an image with the oc installed and some way for it to access the api inside it?
The way I found was to create the Job first and then add the ownerReference in the bc and imageStream.

resources/com/github/coreos/job.json
ravanelli: I think it is just a matter to enable it in the cluster. According to https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/: `this feature is currently alpha, and can be enabled with both kube-apiserver and kube-controller-manager feature gate TTLAfterFinished.`  I was hoping we could maybe open a ticket to the team responsible for the cluster to see if it is possible to be added.

vars/buildImage.groovy
ravanelli: Indeed!!. Much better =)

resources/com/github/coreos/job.json
jlebon: Ahh right OK, that makes total sense now why you had to do instantiate them separately!

resources/com/github/coreos/job.json
jlebon: I'm not sure how easy it is to enable that feature in OCP. We can do some digging, though based on the docs, it seems like it'd require overriding the `kube-apiserver` and `kube-controller-manager` CLIs to have it enabled. I found https://docs.openshift.com/container-platform/4.5/nodes/clusters/nodes-cluster-enabling-features.html related to this, though there's a big caveat that upgrades are not officially supported.

jlebon: So overall, I would probably stick with the pod hack for now.

Or... if we have to run code *anyway* to delete the pod, I wonder if we could just not use ownerReferences at all for now and make the GC code query specifically for BCs which e.g. were created 24h ago?

jlebon: And then in a future release if we have the TTL bit stabilized, we can circle back and clean this up so we don't have to explicitly GC.

vars/buildImage.groovy
ravanelli: I didn't test using a PR. However I'm able to get env.BRANCH_NAME from the branch I'm triggering it. So I get `origin/lib` in my case. Looking at the documentation when it is triggered from a PR the format will be PR-number

vars/buildImage.groovy
ravanelli: I was able to do it for the job template. But for the buildconfig even though I tried the same steps,  some reason the buildoconfig and the imagestream are not created at the end. There isn't any error ou logs to see,  I can only see the parameters is changed ok. Not sure why, probably due the way this template was created, I spent some time on that but couldn't figure out.  I will need to dig more to understand why I can't due it in the second case

resources/com/github/coreos/job.json
ravanelli: So, I basically set 2h time for the job to complete, 2h there  is a plenty of time for the build to complete. After this time the job will be in complete state. The cleanup will look for job names starting with cleanup and will delete these in complete state. If a build fail the job will also end after 2h. Let me know if you want me to change something here =)

resources/com/github/coreos/buildconfig.json
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }
+  ],
+  "parameters": [
+    {
+      "name": "NAME",
+      "displayName": "Name",
+      "description": "The name assigned to all of the frontend objects defined in this template.",
+      "required": true,
+      "value": "cosa-image"
```
jlebon, line 110: Let's remove this one to make sure it has to be specified.

resources/com/github/coreos/buildconfig.json
```diff
@@ -0,0 +1,163 @@
...
+      "description": "The OpenShift Namespace where the ImageStream resides.",
+      "required": true,
+      "value": "coreos-ci"
+    },
+    {
+      "name": "SOURCE_REPOSITORY_URL",
+      "displayName": "Git Repository URL",
+      "description": "The URL of the repository with your application source code.",
+      "required": true,
+      "value": "none"
```
jlebon, line 124: This one too.

resources/com/github/coreos/buildconfig.json
```diff
@@ -0,0 +1,163 @@
...
+    },
+    {
+      "name": "SOURCE_REPOSITORY_REF",
+      "displayName": "Git Reference",
+      "description": "Set this to a branch name, tag or other ref of your repository if you are not using the default branch.",
+      "value": "master"
```
jlebon, line 130: And this one too.

resources/com/github/coreos/job.json
```diff
@@ -0,0 +1,52 @@
...
+    "template": {
+      "spec": {
+        "containers": [
+          {
+            "name": "${NAME}",
+            "image": "${IMAGE}",
```
jlebon, line 26: Let's just hardcode an image here like Busybox or Fedora and drop the `IMAGE` parameter entirely? Basically anything that has `sleep` in it. :)

resources/com/github/coreos/job.json
```diff
@@ -0,0 +1,52 @@
...
+            "command": [
+              "sleep", "2h"
```
jlebon, line 28: Maybe more to give time to debug if needed? 24h seems reasonable.

resources/com/github/coreos/job.json
```diff
@@ -0,0 +1,52 @@
...
+    "name": "clean-up",
+    "annotations": {
+      "openshift.io/display-name": "Clean Up Job",
+      "description": "Clean Up Job",
+      "template.openshift.io/bindable": "false"
+    }
+  },
+  "objects": [
+  {      
+  "apiVersion": "batch/v1",
```
jlebon, line 14: Missing some indentation here. Might be worth passing through `jq` to get it pretty-printed (though as mentioned in the first comment, hopefully we switch to YAML soon! :) )

vars/buildImage.groovy
```diff
@@ -0,0 +1,81 @@
...
+                   error("GitURL is a required field. Aborting")
+               }
+
+               def bcJSON  = libraryResource 'com/github/coreos/buildconfig.json'
+               def jobJSON = libraryResource 'com/github/coreos/job.json'
+               def jobObj  = readJSON text: jobJSON
+               bcObj       = readJSON text: bcJSON
+               def tagName = UUID.randomUUID()
+
+               jobObj      = openshift.process(jobObj, "-p", "NAME=cleanup-${tagName}")
```
jlebon, line 38: Instead of `cleanup-$UUID` and `bc-$UUID`, to be a little more descriptive, how about `coreos-ci-$repo-$UUID`? Where `$repo` is (the Groovy equivalent of) `$(basename $SOURCE_REPOSITORY_URL)`.

OK to defer this too if you prefer (though maybe at least for now `coreos-ci-$UUID`?).

vars/buildImage.groovy
```diff
@@ -0,0 +1,81 @@
...
+               def app     = openshift.create(jobObj)
+               def job     = app.narrow('job')
+               def jobMap  = job.object()
+               def uid     = jobMap.metadata.get('uid')
+
+               imageName = "bc-${tagName}"
+               //Create and unique name for image and bc
+               bcObj['parameters'][0] +=['name': 'NAME', 'value': "bc-${tagName}".toString()]
```
jlebon, line 46: Using indexes like that will be hard to maintain. That'll be good motivation too to tackle https://github.com/coreos/coreos-ci-lib/pull/61#discussion_r569401940 as a follow-up!

resources/com/github/coreos/buildconfig.json
```diff
@@ -0,0 +1,163 @@
...
+        ]
+      },
+      "spec": {
+        "source": {
+          "type": "Git",
+          "git": {
+            "uri": "${SOURCE_REPOSITORY_URL}",
+            "ref": "${SOURCE_REPOSITORY_REF}"
+          }
+        },
```
jlebon, line 69: Hmm actually, we should be able to just drop this and instead do `openshift.startBuild("--from-dir=.")` (could make the dir path overridable via a parameter to `buildImage()`) which will pass the workspace as the build input. The major advantage of this is that it closes a race condition where the ref was moved since the job was triggered. It's also simpler because we don't have to fetch around for env vars and parse it out; basically it puts the onus on the caller to fetch the source. :)

vars/buildImage.groovy
```diff
@@ -0,0 +1,81 @@
+// Create a Job and BuildConfig Template to trigger the Build
```
jlebon, line 1: Maybe a more generic description? E.g. something like:

> Builds a container image and returns its name.

resources/com/github/coreos/buildconfig.json
ravanelli: Thanks again @jlebon \o/,  awesome,  it also solves the issue to get the PR reference. It is much cleaner in this way. 
The only think missing is the stuff related to[ #61](https://github.com/coreos/coreos-ci-lib/pull/61#discussion_r569401940).  I will get some help from the community to understand it.

ravanelli:
Ummm I will need to add /hold for now. I just noticed the openshift_hack is messing the things now using --from-dir. I will need to take a look on that first

jlebon:
> Ok, according to coreos/coreos-assembler@82bc48a. That's the reason it doesn't work. We still need to pass the $GIT_REF and $GIT_URL in order to have the openshift_hack working. Even being in a newer version of ocp it still doesn't create the .git, requiring the hack.

Bahh, that sucks. Hmm... have you tried bundling up the directory in a tarball instead and using `--from-archive=dir.tar` instead? Seems like that should workaround any filtering `oc start-build` might try to do.

jlebon:
> have you tried bundling up the directory in a tarball instead and using --from-archive=dir.tar instead? Seems like that should workaround any filtering oc start-build might try to do.

OK, that doesn't work either locally because OpenShift still filters out `.git`. I've pushed out https://github.com/coreos/coreos-assembler/pull/2043 which works with this now: the cosa CI simply would need to do `tar -cJf git.tar.gz .git` before calling `buildImage()`.

ravanelli:
> OK, that doesn't work either locally because OpenShift still filters out `.git`. I've pushed out [coreos/coreos-assembler#2043](https://github.com/coreos/coreos-assembler/pull/2043) which works with this now: the cosa CI simply would need to do `tar -cJf git.tar.gz .git` before calling `buildImage()`.

How did you test it?
I was able to use .tar, It worked in this way:
` shwrap("tar --exclude=./dir.tar --exclude=./pod* -zcvf dir.tar ${workspace}")
  def bc = openshift.selector('bc', imageName).startBuild("--from-archive=dir.tar")`
  
Not sure if you can see it: 
https://jenkins-rravanel-fcos-dev.apps.ocp.ci.centos.org/job/cosa-build/view/change-requests/job/PR-2-head/21/consoleFull

jlebon:
So here's what I'm trying:

```diff
diff --git a/Dockerfile b/Dockerfile
index ac51610b..a4f1b9b9 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,9 @@
 FROM registry.fedoraproject.org/fedora:33
 WORKDIR /root/containerbuild

+COPY ./ /root/containerbuild/
+RUN ls -la
+
 # We split into multiple steps here so that local dev workflows which involve
 # rebuilding the container can leverage cached layers better.

```

(Just so that I can see earlier if it worked instead of waiting for all the packages to get installed.)

And then:

```
$ tar -czf ../cosa.tar.gz .
```

But the `.git` directory didn't survive the transfer:

```
$ oc start-build jlebon-coreos-assembler --from-archive=cosa.tar.gz --follow
...
Step 3/23 : COPY ./ /root/containerbuild/
 ---> 7b2b6911ecbb
Removing intermediate container e73c2c821c8b
Step 4/23 : RUN ls -la
 ---> Running in 98054eab8071
total 114100
drwxr-xr-x.  1 root root      4096 Feb 16 18:43 .
dr-xr-x---.  1 root root        28 Feb 16 18:43 ..
-rw-rw-r--.  1 root root      3861 Feb 16 18:41 .cci.jenkinsfile
-rw-rw-r--.  1 root root        93 Feb 16 18:41 .dockerignore
drwxrwxr-x.  3 root root        28 Feb 16 18:41 .github
-rw-rw-r--.  1 root root       112 Feb 16 18:41 .gitignore
-rw-rw-r--.  1 root root       123 Feb 16 18:41 .gitmodules
-rw-rw-r--.  1 root root      3863 Feb 16 18:41 CONTRIBUTING.md
-rw-rw-r--.  1 root root       956 Feb 16 18:43 Dockerfile
-rw-rw-r--.  1 root root       689 Feb 16 18:41 Dockerfile.buildroot
-rw-rw-r--.  1 root root       906 Feb 16 18:41 Dockerfile.dev
-rw-rw-r--.  1 root root     11325 Feb 16 18:41 LICENSE
-rw-rw-r--.  1 root root      3280 Feb 16 18:41 Makefile
-rw-rw-r--.  1 root root       216 Feb 16 18:41 OWNERS
-rw-rw-r--.  1 root root      1974 Feb 16 18:41 README.md
-rw-rw-r--.  1 root root       517 Feb 16 18:41 RELEASE.txt
-rwxrwxr-x.  1 root root      2306 Feb 16 18:41 bottlecap
-rwxrwxr-x.  1 root root       174 Feb 16 18:41 bottlecap-shim
-rwxrwxr-x.  1 root root      4326 Feb 16 18:41 build.sh
drwxrwxr-x.  2 root root        48 Feb 16 18:41 ci
-rw-rw-r--.  1 root root      3039 Feb 16 18:41 code-of-conduct.md
drwxrwxr-x.  6 root root      4096 Feb 16 18:41 docs
drwxrwxr-x.  2 root root        57 Feb 16 18:41 exttests
drwxrwxr-x.  2 root root        41 Feb 16 18:41 fixtures
drwxrwxr-x.  8 root root       159 Feb 16 18:41 gangplank
drwxrwxr-x. 19 root root      4096 Feb 16 18:42 mantle
drwxrwxr-x.  2 root root       118 Feb 16 18:42 ocp
-rw-rw-r--.  1 root root       115 Feb 16 18:42 pytest.ini
drwxrwxr-x.  4 root root      4096 Feb 16 18:42 src
-rw-rw-r--.  1 root root 116731085 Feb 16 18:43 tags
-rw-rw-r--.  1 root root        25 Feb 16 18:42 test-requirements.txt
drwxrwxr-x.  3 root root       146 Feb 16 18:42 tests
drwxrwxr-x.  3 root root        97 Feb 16 18:42 tools
 ---> 105410d8dbc4
Removing intermediate container 98054eab8071
```

Even though it's in the archive:

```
$ tar -tf cosa.tar.gz ./.git
./.git/
./.git/hooks/
./.git/hooks/pre-commit
./.git/refs/
./.git/refs/heads/
./.git/refs/heads/master
./.git/refs/tags/
./.git/refs/remotes/
./.git/refs/remotes/origin/
./.git/refs/remotes/origin/HEAD
./.git/HEAD
./.git/config
./.git/objects/
./.git/objects/pack/
./.git/objects/pack/pack-9e0c41873c17de46bfd28009b2ba715676fa2cae.pack
./.git/objects/pack/pack-9e0c41873c17de46bfd28009b2ba715676fa2cae.idx
./.git/objects/info/
./.git/objects/a4/
./.git/objects/a4/f1b9b9a77feacf8eb2ad60370e9bfb5ea57ff4
./.git/objects/04/
./.git/objects/04/c5076aab88a6781dfd3c9c988d461e42f90e8a
./.git/objects/8b/
./.git/objects/8b/e2c398f7d7eb35c220f2a5ccc3ee7484519bc9
./.git/packed-refs
./.git/logs/
./.git/logs/refs/
./.git/logs/refs/remotes/
./.git/logs/refs/remotes/origin/
./.git/logs/refs/remotes/origin/HEAD
./.git/logs/refs/heads/
./.git/logs/refs/heads/master
./.git/logs/HEAD
./.git/index
./.git/COMMIT_EDITMSG
```

ravanelli:
Here is the output from the build from [Jenkins](https://jenkins-rravanel-fcos-dev.apps.ocp.ci.centos.org/job/cosa-build/view/change-requests/job/PR-2-head/21/console) 

```
+ set -xeuo pipefail
++ umask
+ '[' 0022 = 0000 ']'
+ tar --exclude=./dir.tar '--exclude=./pod*' -zcvf dir.tar .
./
./.git/
./.git/branches/
./.git/description
./.git/hooks/
./.git/hooks/applypatch-msg.sample
./.git/hooks/commit-msg.sample
./.git/hooks/fsmonitor-watchman.sample
./.git/hooks/post-update.sample
./.git/hooks/pre-applypatch.sample
./.git/hooks/pre-commit.sample
./.git/hooks/pre-push.sample
./.git/hooks/pre-rebase.sample
./.git/hooks/pre-receive.sample
./.git/hooks/prepare-commit-msg.sample
./.git/hooks/update.sample
./.git/info/
./.git/info/exclude
./.git/refs/
./.git/refs/heads/
./.git/refs/tags/
./.git/refs/remotes/
./.git/refs/remotes/origin/
./.git/refs/remotes/origin/PR-2-head
./.git/HEAD
./.git/config
./.git/objects/
./.git/objects/pack/
./.git/objects/pack/pack-11bc2ef21978cc789758bd1fd8808eccf3aba4e2.pack
./.git/objects/pack/pack-11bc2ef21978cc789758bd1fd8808eccf3aba4e2.idx
./.git/objects/info/
./.git/FETCH_HEAD
./.git/logs/
./.git/logs/refs/
./.git/logs/refs/remotes/
./.git/logs/refs/remotes/origin/
./.git/logs/refs/remotes/origin/PR-2-head
./.git/logs/HEAD
./.git/index
./.cci.jenkinsfile
./.dockerignore
./.github/
./.github/ISSUE_TEMPLATE/
./.github/ISSUE_TEMPLATE/bug.md
./.github/ISSUE_TEMPLATE/feature.md
./.gitignore
./.gitmodules
```

```
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Receiving source from STDIN as archive ...
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Caching blobs under "/var/cache/blobs".
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] 
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Pulling image registry.fedoraproject.org/fedora:33 ...
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Getting image source signatures
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Copying blob sha256:157ab801145489f145f258148bd135102a3294e420f1859a39e824e7cda56b2f
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Copying config sha256:9f2a56037643a68ea81711a8eeb4501428eefd40b000c866ad9745a581c0464d
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Writing manifest to image destination
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Storing signatures
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] Adding transient rw bind mount for /run/secrets/rhsm
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] STEP 1: FROM registry.fedoraproject.org/fedora:33
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] STEP 2: ENV "GIT_URL"="none" "GIT_REF"="none"
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] --> 0096c42a097
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] STEP 3: WORKDIR /root/containerbuild
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] --> 6d3cd5b4e19
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] STEP 4: RUN echo "renata"
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] renata
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] --> f0051fa0748
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] STEP 5: COPY ./src/print-dependencies.sh ./src/deps*.txt ./src/vmdeps*.txt ./src/build-deps.txt /root/containerbuild/src/
```
....
```
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] STEP 9: COPY ./ /root/containerbuild/
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] --> c33bae79e90
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] STEP 10: RUN ./build.sh write_archive_info
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ pwd
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + srcdir=/root/containerbuild
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + '[' 1 -ne 0 ']'
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + write_archive_info
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + . /root/containerbuild/src/cmdlib.sh
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ set -euo pipefail
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] +++ dirname ./build.sh
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ DIR=.
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ RFC3339=%Y-%m-%dT%H:%M:%SZ
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ grep -q '^Fedora' /etc/redhat-release
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ export ISFEDORA=1
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ ISFEDORA=1
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ export ISEL=
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ ISEL=
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] +++ python3 -c '
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] import gi
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] gi.require_version("RpmOstree", "1.0")
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] from gi.repository import RpmOstree
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] print(RpmOstree.get_basearch())'
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ basearch=x86_64
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ export basearch
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] +++ uname -m
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ arch=x86_64
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ export arch
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ case $arch in
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ DEFAULT_TERMINAL=ttyS0
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ export DEFAULT_TERMINAL
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ _privileged=
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + mkdir -p /cosa /lib/coreos-assembler
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + touch -f /lib/coreos-assembler/.clean
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + prepare_git_artifacts /root/containerbuild /cosa/coreos-assembler-git.tar.gz /cosa/coreos-assembler-git.json
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local gitd=/root/containerbuild
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + shift
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local tarball=/cosa/coreos-assembler-git.tar.gz
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + shift
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local json=/cosa/coreos-assembler-git.json
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + shift
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + openshift_git_hack /root/containerbuild
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local gitd=/root/containerbuild
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + shift
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + '[' x == x ']'
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + return
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local is_dirty=false
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local head_ref=unknown
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local head_remote=unknown
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local head_url=unknown
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local 'gc=git --work-tree=/root/containerbuild --git-dir=/root/containerbuild/.git'
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + git --work-tree=/root/containerbuild --git-dir=/root/containerbuild/.git diff --quiet --exit-code
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + is_dirty=true
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + tar -C /root/containerbuild -czf /cosa/coreos-assembler-git.tar.gz --exclude-vcs .
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + chmod 0444 /cosa/coreos-assembler-git.tar.gz
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local rev
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local branch
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ git --work-tree=/root/containerbuild --git-dir=/root/containerbuild/.git rev-parse HEAD
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + rev=b6018faa2354f1d6d83cce66fb0b6a74e8e967a0
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ git --work-tree=/root/containerbuild --git-dir=/root/containerbuild/.git rev-parse --abbrev-ref HEAD
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + branch=HEAD
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + [[ -f /root/containerbuild/.git/shallow ]]
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + [[ HEAD == \H\E\A\D ]]
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ git --work-tree=/root/containerbuild --git-dir=/root/containerbuild/.git remote get-url origin
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + head_url=https://github.com/ravanelli/coreos.git
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + info 'Directory /root/containerbuild, is from branch HEAD, commit b6018faa2354f1d6d83cce66fb0b6a74e8e967a0'
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + echo 'info: Directory /root/containerbuild, is from branch HEAD, commit b6018faa2354f1d6d83cce66fb0b6a74e8e967a0'
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] info: Directory /root/containerbuild, is from branch HEAD, commit b6018faa2354f1d6d83cce66fb0b6a74e8e967a0
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + local checksum name size
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ sha256sum /cosa/coreos-assembler-git.tar.gz
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ awk '{print$1}'
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + checksum=b0d7e3dbbb141b9c70036efde09f1d7a0f592933d6297e31750b885f4fa84197
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ basename /cosa/coreos-assembler-git.tar.gz
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + name=coreos-assembler-git.tar.gz
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ find /cosa/coreos-assembler-git.tar.gz -printf %s
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + size=27989939
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + cat
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] ++ date -u +%Y-%m-%dT%H:%M:%SZ
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] + chmod 0444 /cosa/coreos-assembler-git.json
[logs:build/coreos-ci-coreos-aaa60cf8-970b-43f4-afae-d1de3cf33591-1] --> 9bb545652ef
```

> So here's what I'm trying:
> +COPY ./ /root/containerbuild/
> +RUN ls -la

A nice,  Let me add the same and run it again to make sure it is really getting the .git

ravanelli:
The centOs cluster is running now in 4.6. In what version are you?

So it is getting the .git : 
```
STEP 6: RUN ls -la
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] total 108
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] drwxr-xr-x.  1 root root  4096 Feb 16 19:42 .
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] dr-xr-x---.  1 root root    28 Feb 16 19:42 ..
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] -rw-r--r--.  1 root root  3861 Feb 16 19:41 .cci.jenkinsfile
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] -rw-r--r--.  1 root root    93 Feb 16 19:41 .dockerignore
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] drwxr-xr-x.  8 root root   162 Feb 16 19:41 .git
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] drwxr-xr-x.  3 root root    28 Feb 16 19:41 .github
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] -rw-r--r--.  1 root root   112 Feb 16 19:41 .gitignore
[logs:build/coreos-ci-coreos-5bc8913e-e289-4a45-bdd3-ed5d7e596bde-1] -rw-r--r--.  1 root root   123 Feb 16 19:41 .gitmodules
```

jlebon:
Ahh sweet, yes this does look like it's sensitive to the OCP version.
I was using v3.11 (the internal PSI cluster), but indeed in CentOS CI it works.
So this was fixed at some point in OCP 4.6.

For completeness, I also tested `--from-repo=.` and `--from-dir=.` and both of them still filter `.git`. But at least `--from-archive` works.

resources/com/github/coreos/buildconfig.json
```diff
@@ -0,0 +1,137 @@
...
+      "description":"The URL of the repository with your application source code.",
+      "required":true,
+      "value":"none"
+    },
+    {
+      "name":"SOURCE_REPOSITORY_REF",
+      "displayName":"Git Reference",
+      "description":"Set this to a branch name, tag or other ref of your repository if you are not using the default branch.",
+      "value":"none"
+    },
```
jlebon, line 111: We shouldn't need those parameters anymore, right?

resources/com/github/coreos/job.yaml
```diff
@@ -0,0 +1,26 @@
...
+          command:
+          - sleep
+          - 2h
+        restartPolicy: Never
+parameters:
+- name: NAME
+  required: true
+  value: cleanupJob
+- name: IMAGE
+  value: fedora:33
```
jlebon, line 26: Can just drop this parameter and hard-code it into the job itself.

vars/buildImage.groovy
```diff
@@ -0,0 +1,74 @@
+// Builds a container image and returns its name.
+// Available parameters:
+//    dockerFile    string -- DockerFile used for the buildconfig
+//    nameSpace     string -- Namespace name to run the build
```
jlebon, line 4: Does it even make sense to support this as a parameter? I don't think we have the ACLs to create objects in any other namespace anyway.

jlebon (commented):
Just a few final comments, but this is looking really close!

jlebon (approved):
Nice work! :tada:

PR #62, opened 2021-02-03
Automatically set MAKE_JOBS

cgwalters:
See https://github.com/coreos/coreos-ci/issues/23
We're doing this manually in rpm-ostree CI; let's standardize
on this.

### DIFF
--- a/vars/pod.groovy
+++ b/vars/pod.groovy
@@ -27,6 +27,9 @@ def call(params = [:], Closure body) {
     }
     if (params['cpu']) {
         podObj['spec']['containers'][0]['resources']['requests']['cpu'] = params['cpu'].toString()
+        // Also propagate CPU count to NCPUS, because it can be hard in a Kubernetes environment
+        // to determine how much CPU one should really use.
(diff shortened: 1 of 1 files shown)


vars/pod.groovy
```diff
@@ -27,6 +27,9 @@ def call(params = [:], Closure body) {
     }
     if (params['cpu']) {
         podObj['spec']['containers'][0]['resources']['requests']['cpu'] = params['cpu'].toString()
+        // Also propagate CPU count to MAKE_JOBS, because it can be hard in a Kubernetes environment
+        // to determine how much CPU one should really use.
+        podObj['spec']['containers'][0]['env'] += ['name': 'MAKE_JOBS', 'value': params['cpu'].toString()]
```
jlebon, line 32: WDYT about `NCPUS` instead to make it more builder agnostic?

cgwalters: Sure, done.

jlebon (approved):
Good idea, thanks!

PR #64, opened 2021-03-18
utils: add pathExists

jlebon:
Upstreamed from
https://github.com/coreos/fedora-coreos-pipeline/blob/5cf784ba931627c1c08dbb9f09080c57542480c9/utils.groovy#L63.


jlebon:
Ahh right, totally forgot that GitHub auto-requests reviews from `CODEOWNERS`. Sorry for the triple review requests!

ravanelli (approved):
/lgtm

Issue #63, opened 2021-03-17
buildImage() doesn't fail if image building fails


jlebon:
See https://github.com/coreos/coreos-assembler/pull/2083#issuecomment-801198275.

ravanelli:
See it was already fixed in https://github.com/coreos/coreos-ci-lib/pull/65. Closing it

PR #65, opened 2021-03-19
Issue #63 - Fix buildImage

ravanelli:
- Fail buildImage if image build fails

### DIFF
--- a/vars/buildImage.groovy
+++ b/vars/buildImage.groovy
@@ -58,11 +58,22 @@ def call(params = [:]) {
             stage('Build') {
                 def workspace = params.get('workspace', ".");
                 shwrap("tar --exclude=./pod* -zcf /tmp/dir.tar ${workspace}")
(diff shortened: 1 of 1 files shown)


vars/buildImage.groovy
```diff
@@ -63,6 +63,17 @@ def call(params = [:]) {
                 // Showing logs in Jenkins is also a way
                 // to wait for the build to finsih
                 bc.logs('-f')
```
jlebon, line 65: Ahh OK, one thing that tripped me up here is that `startBuild` returns a "build selector" which selects the build which was started. The `bc` makes it seem like it's the `buildConfig` object itself.

Maybe let's rename that var to `build`?

jlebon (approved):
Thanks for tackling this quickly! Had one comment, but LGTM otherwise! Were you able to test the failure case?

vars/buildImage.groovy
ravanelli: Thanks for the review @jlebon. Yes, I was able to test it forcing a synthesis error in cosa.
Done, changed it to build =)

PR #66, opened 2021-03-25
Drop support for `cosaPod(buildroot: true)`; add `buildPod`

jlebon:
We're moving away from a buildroot image based on cosa to one based on
FCOS:

https://github.com/coreos/fedora-coreos-config/pull/740

Correspondingly, add a new `buildPod` as a thin sugar for selecting that
image.


jlebon:
You can see what this looks like in https://github.com/coreos/ignition/pull/1182.

cgwalters (approved):

PR #69, opened 2021-03-29
buildPod: use dumb-init by default

jlebon:
Otherwise, we risk running out of resources in our pods.


jlebon:
Requires: https://github.com/coreos/fedora-coreos-config/pull/916

PR #71, opened 2021-04-29
Add gangplank helper

ravanelli:
- Add a helper to facilitate gangplank calls

### DIFF
--- a/vars/gangplank.groovy
+++ b/vars/gangplank.groovy
@@ -1,0 +1,143 @@
+// Run gangplank
+
+// Available parameters:
+//     artifacts:    list    -- list of artifacts to be built
+//     extraFlags    string  -- Extra flags to use
+//     image:        string  -- Name of the image to be used
+//     mode:         string  -- Gangplank Mode
+//     workDir:      string  -- cosa working directory
+
+def buildArtifact(params = [:]) {
+    gangplankCmd = getMode(params)
+    gangplankCmd = getArtifacts(params, gangplankCmd)
+    runGangplank(gangplankCmd)
+}
+
+// Available parameters:
+//     artifacts:    list    -- list of artifacts to be built
+//     extraFlags    string  -- Extra flags to use
+//     image:        string  -- Name of the image to be used
+//     mode:         string  -- Gangplank Mode
+//     minioServeDir string  -- location to service minio from
+//     workDir:      string  -- cosa working directory
+
+def buildParallelArtifacts(params = [:]) {
+    gangplankCmd = getMode(params)
+    if (params['artifacts']) {
+        def artifacts = params['artifacts']
+        def configFile = startMinio(params)
+
+        // Call one pod for each artifact in parallel
+        parallel artifacts.inject([:]) { d, i -> d[i] = {
+            startParallel(configFile, "${gangplankCmd} -A ${i} ")
+        }; d }
+        finalize()
+    }
+}
+
+// Available parameters:
+//     artifacts:    list    -- list of artifacts to be built
+//     extraFlags    string  -- Extra flags to use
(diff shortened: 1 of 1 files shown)


vars/gangplank.groovy
```diff
@@ -0,0 +1,87 @@
+// Run gangplank
+def ganplankCmd
```
jlebon, line 2: Hmm, seems like we could pretty easily get rid of this global var by having `getMode()` return it and then passing it in to the various helpers and then finally `runGangplank()`.

vars/gangplank.groovy
```diff
@@ -0,0 +1,87 @@
...
+
+def buildArtifact(params = [:]) {
+// Available parameters:
```
jlebon, line 5: For consistency with the other groovy commands, let's put the docstring above the functions?

vars/gangplank.groovy
```diff
@@ -0,0 +1,87 @@
...
+//     artifacts:    list    -- list of artifacts to be built
+//     extraFlags    list    -- Extra flags to use
+//     image:        string  -- Name of the image to be used
+//     workDir:      string  -- cosa working directory
```
jlebon, line 9: Missing `mode`

vars/gangplank.groovy
```diff
@@ -0,0 +1,87 @@
...
+        currentBuild.result = 'ABORTED'
+        error("Error running gangplank.")
+    }
+}
+
+def runSpec(params =[:]) {
+// Available parameters:
+//     artifacts:    list    -- list of artifacts to be built
+//     extraFlags    list    -- Extra flags to use
+//     spec          string  -- Name of the spec file
```
jlebon, line 48: Missing `mode` here too.

darkmuggle (commented):
For the RHCOS and Developer use-cases we'll need to support `--git-{ref,url}` and `--repo` as well -- and then we can likely start a parallel RHCOS pipeline using the FCOS pipeline.

ravanelli:
> For the RHCOS and Developer use-cases we'll need to support `--git-{ref,url}` and `--repo` as well -- and then we can likely start a parallel RHCOS pipeline using the FCOS pipeline.

For this case, there is the extraFlags you can pass whatever you want. But let me know if you think it would be better adding it explicitly.

darkmuggle:
This is good and I think we can merge as is but we'll need one more command to run _after_ all the gangplank have come back: `cosa meta --finalize` or `gangplank pod -A finalize`  to bring all the `meta.*.json` files into a single `meta.json`.

darkmuggle:
/lgtm

Great work @ravanelli!

vars/gangplank.groovy
```diff
@@ -0,0 +1,143 @@
...
+    runGangplank(gangplankCmd)
+    return fileName
+}
+
+def startParallel(configFile, gangplankCmd) {
+    gangplankCmd += "-m ${configFile}"
+    runGangplank(gangplankCmd)
+}
+
+def startMinio(params =[:]) {
```
jlebon, line 62: I think basically every function which takes the generic `params` arg and is intended to be "user-facing" should have a docstring with supported params. We can add one to this one in a follow-up! :)

vars/gangplank.groovy
```diff
@@ -0,0 +1,143 @@
...
+    return gangplankCmd
+}
+
+def getWorkDir(params = [:], gangplankCmd) {
+    if(params['cosaDir']) {
+        gangplankCmd += "--workDir ${utils.getCosaDir(params)} "
+        return gangplankCmd
+    }
+    return gangplankCmd
+}
```
jlebon, line 142: If these are more implementation details, I wonder if we should prefix them with e.g. `_` instead. WDYT? Not a blocker.

PR #72, opened 2021-05-04
move to f34 for the build image

dustymabe:


resources/com/github/coreos/job.yaml
```diff
@@ -13,7 +13,7 @@ objects:
       spec:
         containers:
         - name: "${NAME}"
-          image: fedora:33
```
dustymabe, line 16: should we be using `registry.fedoraproject.org/fedora:34` here?

ravanelli: We can move it to 34, but It won't affect the cosa build itself, since it uses the [docker](https://github.com/coreos/coreos-assembler/blob/master/Dockerfile) file from coreos-assembler repo in the [bc](https://github.com/coreos/coreos-ci-lib/blob/a062fd9b8c11dc1a3a73cb82cad1c34cc6534289/resources/com/github/coreos/buildconfig.json#L60).

 The Fedora:33 in this case is just to start a Job in the cluster, in order to create a dependency to clean up the past builds/images ;)

ravanelli: But yeah, since F34 is there,  better move forward.

ravanelli (approved):

resources/com/github/coreos/job.yaml
jlebon: Using the Fedora registry SGTM if you'd like to change that. Honestly, I'm sure there's a super tiny image even smaller than e.g. busybox we could use for this.

jlebon (approved):

dustymabe:
Updated to use the fedora registry

jlebon:
Ahhh, I should've asked you to tweak the commit title too. Wrote a docs PR to repent: https://github.com/coreos/coreos-ci-lib/pull/73

PR #75, opened 2021-05-12
buildImage: use untilEach() instead of watch() and add timeout

jlebon:
The way we use `watch`, the closure only runs once and then exits. But
we want to keep watching for changes until the build goes to `Complete`.
Instead, just use `untilEach()` which is nicer for this. This mirrors
the last example in:

https://github.com/openshift/jenkins-client-plugin#watching-and-waiting-of-course

Also add a build timeout of 2 hours to be safe.

### DIFF
--- a/vars/buildImage.groovy
+++ b/vars/buildImage.groovy
@@ -63,16 +63,18 @@ def call(params = [:]) {
                 // Showing logs in Jenkins is also a way
                 // to wait for the build to finsih
                 build.logs('-f')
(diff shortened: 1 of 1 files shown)


ravanelli (approved):

jlebon:
Hmm though if I understand correctly, this will dumbly wait 2h if the build fails, which isn't what we want either. Let me tweak this.

jlebon:
OK, updated now!

ravanelli (approved):

PR #77, opened 2021-05-13
pod: Add param sa (serviceAccount) in pod

ravanelli:
- In order to run gangplank we need to use jenkins
 service account, let's allow this option in pod

### DIFF
--- a/vars/pod.groovy
+++ b/vars/pod.groovy
@@ -7,12 +7,17 @@
 //   cpu: amount of CPU to request
 //   emptyDirs: []string
 //   cmd: []string
+//   serviceAccount: string
 //   secrets: []string
 //   configMaps: []string
 def call(params = [:], Closure body) {
     def podJSON = libraryResource 'com/github/coreos/pod.json'
     def podObj = readJSON text: podJSON
     podObj['spec']['containers'][0]['image'] = params['image']
+
+    if (params['serviceAccount'] != null) {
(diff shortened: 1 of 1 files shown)


vars/pod.groovy
```diff
@@ -7,13 +7,18 @@
 //   cpu: amount of CPU to request
 //   emptyDirs: []string
 //   cmd: []string
+//   sa: string
```
jlebon, line 10: Let's just spell this out (to `serviceAccount`)?

jlebon (approved):
> In order to run gangplank we need to use jenkins service account

Hmm interesting. I would've thought everything created by Jenkins already used the Jenkins service account. But indeed, checking a random pod that was already running, it's actually using the `default` SA. Seems like we should just always use the `jenkins` SA instead? Anyway, we can look into that later.

One nit, LGTM otherwise!

ravanelli:
> > In order to run gangplank we need to use jenkins service account
> 
> Hmm interesting. I would've thought everything created by Jenkins already used the Jenkins service account. But indeed, checking a random pod that was already running, it's actually using the `default` SA. Seems like we should just always use the `jenkins` SA instead? Anyway, we can look into that later.
> 
> One nit, LGTM otherwise!

Just the gangplank use yes, we would always use it. But I thought about leaving a flexible option, for any sa name.

PR #78, opened 2021-05-13
Fix serviceAccount declaration in pod

ravanelli:


ravanelli:
My bad for this one. I was testing it with more permission in the default sa, it allows me to run it. Turns out the sa shouldn't be part of the container declaration.

jlebon (approved):

PR #81, opened 2021-06-10
fcosBuild: Fix branch reference

ravanelli:
- It will always point to the main branch even
if we are running the CI for other branches.
For this case we need to match the fedora-coreos-config
branch with the CI branch running. So, let's the right branch be
determined by who is calling the function.

### DIFF
--- a/vars/fcosBuild.groovy
+++ b/vars/fcosBuild.groovy
@@ -1,15 +1,16 @@
 // Build FCOS, possibly with modifications.
 // Available parameters:
-//    make:           boolean -- run `make && make install DESTDIR=...`
-//    skipInit:       boolean -- assume `cosa init` has already been run
-//    makeDirs:     []string  -- extra list of directories from which to `make && make install DESTDIR=...`
-//    skipKola:       boolean -- don't automatically run kola on resulting build
-//    overlays:     []string  -- list of directories to overlay
-//    extraFetchArgs: string  -- extra arguments to pass to `cosa fetch`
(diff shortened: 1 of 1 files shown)


ravanelli:
Ok, let's try that again. It will also required a backport in cosa to include it in the past branches

vars/fcosBuild.groovy
```diff
@@ -19,7 +20,11 @@ def call(params = [:]) {
         shwrap("mkdir -p ${cosaDir}")
 
         if (!params['skipInit']) {
-            shwrap("cd ${cosaDir} && cosa init https://github.com/coreos/fedora-coreos-config")
+            shwrap("cd ${cosaDir} && git clone https://github.com/coreos/fedora-coreos-config src/config")
+            if (params['gitBranch']) {
+                shwrap("pushd ${cosaDir}/src/config && git checkout origin/${params['gitBranch']}")
+            }
+            shwrap("cd ${cosaDir} && cosa init --force ${cosaDir}/src/config")
```
jlebon, line 27: WDYT of something like this:

```suggestion
            def branchArg = ""
            if (params['gitBranch']) {
                branchArg = "--branch ${params['gitBranch']}"
            }
            shwrap("cd ${cosaDir} && cosa init ${branchArg} https://github.com/coreos/fedora-coreos-config")
```

?

ravanelli: Much better =) thanks @jlebon

jlebon (approved):
Thanks!

PR #83, opened 2021-07-27
gangplank: additions to aid in FCOS AARCH64 building

dustymabe:
Adds runSingleCmd and error checking for runSpec. Also adds
`getGangPlankFCOSAARCH64Params()` with a default set of params for
running against the FCOS AARCH64 builder and
`gangPlankFCOSAARCH64BuilderWrapper()` with default setup and
invocation for executing gangplank against that builder.

### DIFF
--- a/vars/gangplank.groovy
+++ b/vars/gangplank.groovy
@@ -10,7 +10,7 @@
 def buildArtifact(params = [:]) {
     gangplankCmd = _getMode(params)
     gangplankCmd = _getArtifacts(params, gangplankCmd)
-    runGangplank(gangplankCmd)
+    _runGangplank(gangplankCmd)
 }
 
 // Available parameters:
@@ -50,7 +50,7 @@ def generateSpec(params = [:]) {
     gangplankCmd = _getFlags(params, gangplankCmd)
     fileName = "/tmp/jobspec-${UUID.randomUUID()}.spec"
     gangplankCmd += "--yaml-out ${fileName} "
-    runGangplank(gangplankCmd)
+    _runGangplank(gangplankCmd)
     return fileName
 }
 
@@ -59,7 +59,7 @@ def generateSpec(params = [:]) {
 //     gangplankCmd: string  -- Gangplank command to run
 def startParallel(configFile, gangplankCmd) {
     gangplankCmd += "-m ${configFile}"
-    runGangplank(gangplankCmd)
+    _runGangplank(gangplankCmd)
 }
 
 // Available parameters:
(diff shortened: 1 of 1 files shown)


darkmuggle (approved):
/lgtm

Glad to see this happening :)

ravanelli (approved):
/lgtm

Thanks for adding it Dusty!

vars/gangplank.groovy
```diff
@@ -103,6 +103,60 @@ def runSpec(params =[:]) {
...
+// A set of default parameters for building our FCOS AARCH64 builds via gangplank.
+// - use a locally build COSA because we don't currently build/push multi-arch COSA anywhere. 
+// - use pod mode with --podman for remote podman execution
+// - use the 'builds' "bucket" will fetch artifacts back into the expected 'builds' directory
+// - target aarch64 architecture
+def getGangPlankFCOSAARCH64Params() {
+    return [mode: 'pod',
+            extraFlags: '--podman --bucket=builds --arch=aarch64',
+            image: 'localhost/coreos-assembler:latest']
+}
```
jlebon, line 134: Why not fold this into `gangPlankFCOSAARCH64BuilderWrapper`? E.g. we check if `mode` is provided, otherwise default to `pod`, check for `image`, otherwise default to `localhost/coreos-assembler:latest`, etc... For flags, we can have an `appendFlags` or something if we want a pure append and not replacement. See this pattern in e.g. `cosaPod`.

vars/gangplank.groovy
```diff
@@ -103,6 +103,60 @@ def runSpec(params =[:]) {
...
+def gangPlankFCOSAARCH64BuilderWrapper(params = [:]) {
+    withCredentials([
+        string(credentialsId: 'fcos-aarch64-builder-host-string',
+               variable: 'REMOTEHOST'),
+        string(credentialsId: 'fcos-aarch64-builder-uid-string',
+               variable: 'REMOTEUID'),
+        sshUserPrivateKey(credentialsId: 'fcos-aarch64-builder-sshkey-key',
+                          usernameVariable: 'REMOTEUSER',
+                          keyFileVariable: 'CONTAINER_SSHKEY')
+    ]) {
```
jlebon, line 148: Hmm, this seems odd in the library. It's specific to the pipeline. I'd personally prefer having it live in the pipeline as a separate function (and then here we can just pass in the user, host, and uid bits). At the very least, let's make them parameters instead of hardcoding the ID strings.

vars/gangplank.groovy
dustymabe: ok I'll see what I can do to merge this back in.

dustymabe: moved functions to pipeline `gp.groovy`.

vars/gangplank.groovy
dustymabe: moved bits to the pipeline in `gp.groovy`.

dustymabe:
The pipeline specific bits have been moved to the pipeline. Updated list of things done here:

```
commit 5820f10399fa7edddd144547c0e94fc3415f2ccb
Author: Dusty Mabe <dusty@dustymabe.com>
Date:   Tue Jul 27 11:28:37 2021 -0400

    gangplank: additions to aid in FCOS AARCH64 building
    
    - Adds `runSingleCmd()` for running with `--singleCmd`
    - Adds error checking for `runSpec()`.
    - Converts existing `runGangplank()` into internal function `_runGangplank()`
    - Adds `runGangplank()`, switches between `runSingleCmd()` and `runSpec()`
    - Adds `_getArch()` for parsing out and adding `--arch=$arch`
```

jlebon (approved):

PR #84, opened 2021-08-23
fcosKola: support an `arch` parameter for target architecture

dustymabe:
This is useful when we're running cloud tests against a different
architecture than the cosa environment we're running in. For example,
running AWS aarch64 AMI tests.

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -14,6 +14,10 @@ def call(params = [:]) {
     // this is shared between `kola run` and `kola run-upgrade`
     def platformArgs = params.get('platformArgs', "");
     def buildID = params.get('build', "latest");
+    def arch = params.get('arch', "");
+    if (arch != "") {
+        arch = "--arch=${arch}"
+    }
 
     // This is a bit obscure; what we're doing here is building a map of "name"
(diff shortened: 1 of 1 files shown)


ravanelli (approved):
/lgtm

vars/fcosKola.groovy
```diff
@@ -14,6 +14,7 @@ def call(params = [:]) {
     // this is shared between `kola run` and `kola run-upgrade`
     def platformArgs = params.get('platformArgs', "");
     def buildID = params.get('build', "latest");
+    def arch = params.get('arch', "x86_64");
```
jlebon, line 17: Hmm instead of hardcoding x86_64, why not just leave it empty and not pass any `arch` arg at all. E.g.

```suggestion
    def arch = params.get('arch', "");
    if (arch != "") {
        arch = "--arch=${arch}"
    }
```

That way `kola` should default to the arch we're running on, which seems more proper. But OK as is too, we won't have any Jenkins using this code on other arches anytime soon.

jlebon (approved):

dustymabe:
pushed a fixup - now testing

jlebon (approved):

jlebon:
(Assuming the commits will be squashed.)

PR #85, opened 2021-08-27
gangplank: Add bucket option to gangplank

ravanelli:

### DIFF
--- a/vars/gangplank.groovy
+++ b/vars/gangplank.groovy
@@ -2,6 +2,7 @@
 
 // Available parameters:
 //     artifacts:    list    -- list of artifacts to be built
+//     bucket:       string  -- name other minio bucket to use
 //     extraFlags:   string  -- Extra flags to use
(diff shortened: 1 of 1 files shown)


saqibali-2k (approved):

gursewak1997 (approved):

vars/gangplank.groovy
```diff
@@ -146,6 +147,14 @@ def _getArtifacts(params =[:], gangplankCmd) {
     return gangplankCmd
 }
 
+def _getBucket(params =[:], gangplankCmd) {
+     if (params['bucket']) {
+        gangplankCmd += "--bucket ${params['bucket']}"
```
jlebon, line 152: Is this missing a trailing space for safe concatenation in the future?

ravanelli: @jlebon Welcome back from PTO =) Good catch! Thanks for checking it, here is the fix #90

PR #86, opened 2021-08-27
gangplank: Fix finalize call

ravanelli:
- We need to pass the image name and the bucket it is using

### DIFF
--- a/vars/gangplank.groovy
+++ b/vars/gangplank.groovy
@@ -31,7 +31,7 @@ def buildParallelArtifacts(params = [:]) {
         parallel artifacts.inject([:]) { d, i -> d[i] = {
             startParallel(configFile, "${gangplankCmd} -A ${i} ")
         }; d }
-        _finalize()
+        _finalize(params)
     }
 }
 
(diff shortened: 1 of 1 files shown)


vars/gangplank.groovy
```diff
@@ -133,8 +133,12 @@ def runGangplank(params = [:]) {
     }
 }
 
-def _finalize() {
-    _runGangplank("gangplank pod -A finalize")
+def _finalize(params = [:]) {
+    gangplankCmd = "gangplank pod "
+    gangplankCmd = _getImage(params, gangplankCmd)
+    gangplankCmd = _getBucket(params, gangplankCmd)
```
saqibali-2k, line 139: This stood out to me, should this be  `+=` instead of `=`. I'm not too familiar with groovy so correct me if I'm wrong.

saqibali-2k: Oh nvm I see how it works now.

saqibali-2k (approved):

gursewak1997 (approved):

PR #88, opened 2021-09-07
fcosKolaTestIso: also test on UEFI, UEFI + Secure Boot, multipath

bgilbert:
In each case, run a minimal test set by default.

Untested.


jlebon (approved):
Nice! Had a concern about memory, but your last commit addressed that. :)

jlebon:
Was debating in my head whether it'd be better to default to running all the variants here and have projects which don't care opt out, or the other way around. I think doing it this way makes sense, but we should expect some projects to soon want to disable some of the new scenarios in face of slower CI.

bgilbert:
I'm intentionally going around and switching everything to opt-out.  We currently have a patchwork of ISO tests in the various repos, and cases are getting missed.

PR #89, opened 2021-09-07
fcosKolaTestIso: fix name of `iso-as-disk` scenario

bgilbert:
Fixes https://github.com/coreos/coreos-ci-lib/pull/88.


jlebon (approved):
Ironically, we don't have CI set up on this repo yet (https://github.com/coreos/coreos-ci-lib/pull/15).

PR #91, opened 2021-09-09
fcosKola: run basic scenarios by default

bgilbert:
They provide basic coverage of several boot environments that we may not otherwise test.

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -1,13 +1,13 @@
 // Run kola tests on the latest build in the cosa dir
 // Available parameters:
-//    addExtTests:    []string -- list of test paths to run
-//    cosaDir:         string  -- cosa working directory
-//    parallel:        integer -- number of tests to run in parallel (default: 8)
-//    skipUpgrade:     boolean -- skip running `cosa kola --upgrades`
-//    build:           string  -- cosa build ID to target
-//    platformArgs:    string  -- platform-specific kola args (e.g. '-p aws --aws-ami ...`)
-//    extraArgs:       string  -- additional kola args for `kola run` (e.g. `ext.*`)
-//    basicScenarios   boolean -- run basic qemu scenarios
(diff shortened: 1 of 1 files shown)


ravanelli (approved):
LGTM

vars/fcosKola.groovy
```diff
@@ -46,7 +46,7 @@ def call(params = [:]) {
         }
 
         try {
-            if (params['basicScenarios']) {
+            if (!params['skipBasicScenarios']) {
```
ravanelli, line 49: We will also need to change other projects using it as cosa:

https://github.com/coreos/coreos-assembler/blob/ee62041f32b8f916a5f7b3b2ac0143de381e9f12/.cci.jenkinsfile#L22

bgilbert: https://github.com/coreos/coreos-assembler/pull/2422 covers cosa, and I didn't find any others that enable the old flag.

ravanelli: Thanks @bgilbert. I'm not aware of other place, maybe @jlebon or @dustymabe can confirm it.

jlebon: It's in f-c-c too -- I think we'd just need to revert https://github.com/coreos/fedora-coreos-config/commit/82859cbc47bdd0b13d61583431f82fb9456dfb49.

jlebon (approved):
Thanks!

vars/fcosKola.groovy
bgilbert: Whoops, my branch wasn't updated when I grepped. :+1:

bgilbert: https://github.com/coreos/fedora-coreos-config/pull/1209

dustymabe:
some fallout.. suggested fix:  https://github.com/coreos/fedora-coreos-pipeline/pull/393

PR #94, opened 2021-11-24
fcosKola: add disableRerun option

saqibali-2k:
Reruns are off by default in kola. Lets enable them by default here and add an  
option to disable them

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -8,6 +8,7 @@
 //    build:             string   -- cosa build ID to target
 //    platformArgs:      string   -- platform-specific kola args (e.g. '-p aws --aws-ami ...`)
 //    extraArgs:         string   -- additional kola args for `kola run` (e.g. `ext.*`)
+//    disableRerun:      boolean  -- disable reruns of failed tests
 def call(params = [:]) {
     def cosaDir = utils.getCosaDir(params)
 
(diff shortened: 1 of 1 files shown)


vars/fcosKola.groovy
```diff
@@ -8,13 +8,18 @@
...
+//    rerun:             boolean  -- rerun failed tests once
 def call(params = [:]) {
     def cosaDir = utils.getCosaDir(params)
 
     // this is shared between `kola run` and `kola run-upgrade`
     def platformArgs = params.get('platformArgs', "");
     def buildID = params.get('build', "latest");
     def arch = params.get('arch', "");
+    def rerun = ""
+    if (params.get('rerun', false)) {
```
jlebon, line 20: Hmm, don't we want to default to rerunning for CI though? So then this would default to `true`, though maybe it's clearer to make the option `disableRerun` instead (and default to `false`).

saqibali-2k: I was going to make another PR to add `rerun: true` to various calls, but I'm fine with changing this to `disableRerun` as well.

vars/fcosKola.groovy
```diff
@@ -8,13 +8,18 @@
 //    build:             string   -- cosa build ID to target
 //    platformArgs:      string   -- platform-specific kola args (e.g. '-p aws --aws-ami ...`)
 //    extraArgs:         string   -- additional kola args for `kola run` (e.g. `ext.*`)
+//    rerun:             boolean  -- rerun failed tests once
```
jlebon, line 11: This needs to be updated.

jlebon (commented):
Commit message still says `rerun`.

dustymabe:
I think we just need to update the commit message to say `add disableRerun option`, right?

dustymabe (approved):
LGTM

PR #98, opened 2022-01-07
fcosKola: use a unique output directory for each run

dustymabe:
If we try to run mutliple fcosKola() invocations in parallel they
step on each others output directory. Let's write somewhere unique.

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -24,6 +24,9 @@ def call(params = [:]) {
         arch = "--arch=${arch}"
     }
 
+    // Create a unique output directory for this run of fcosKola
+    def outputDir = shwrapCapture("mktemp -d ${cosaDir}/tmp/fcosKola-XXXXX")
+
     // This is a bit obscure; what we're doing here is building a map of "name"
     // to "closure" which `parallel` will run in parallel. That way, we can
     // conditionally only add the `run_upgrades` stage if not explicitly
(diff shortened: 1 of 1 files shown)


vars/fcosKola.groovy
```diff
@@ -24,6 +24,9 @@ def call(params = [:]) {
         arch = "--arch=${arch}"
     }
 
+    // Create a unique output directory for this run of fcosKola
+    outputDir = shwrapCapture("mktemp -d ${cosaDir}/fcosKola-XXXXX")
```
dustymabe, line 28: do we need to worry about cleaning this up at the end?

jlebon: Maybe

```suggestion
    outputDir = shwrapCapture("mktemp -d ${cosaDir}/tmp/fcosKola-XXXXX")
```

instead?

> do we need to worry about cleaning this up at the end?

I think if we put it in the cosa dir's `tmp/` it should be fine to leave around.

dustymabe: Done

jlebon (approved):

vars/fcosKola.groovy
```diff
@@ -24,6 +24,9 @@ def call(params = [:]) {
         arch = "--arch=${arch}"
     }
 
+    // Create a unique output directory for this run of fcosKola
+    outputDir = shwrapCapture("mktemp -d ${cosaDir}/tmp/fcosKola-XXXXX")
```
jlebon, line 28:
```suggestion
    def outputDir = shwrapCapture("mktemp -d ${cosaDir}/tmp/fcosKola-XXXXX")
```

jlebon (changes requested):
Wait, sorry.

vars/fcosKola.groovy
```diff
@@ -52,26 +55,26 @@ def call(params = [:]) {
...
-            shwrap("cd ${cosaDir} && cosa kola run ${rerun} --build=${buildID} ${arch} ${platformArgs} --parallel ${parallel} ${args} ${extraArgs}")
+            shwrap("cd ${cosaDir} && cosa kola run ${rerun} --output-dir=${outputDir} --build=${buildID} ${arch} ${platformArgs} --parallel ${parallel} ${args} ${extraArgs}")
         } finally {
-            shwrap("tar -c -C ${cosaDir}/tmp kola | xz -c9 > ${env.WORKSPACE}/kola.tar.xz")
+            shwrap("tar -c -C ${outputDir} kola | xz -c9 > ${env.WORKSPACE}/kola.tar.xz")
             archiveArtifacts allowEmptyArchive: true, artifacts: 'kola.tar.xz'
         }
         // sanity check kola actually ran and dumped its output in tmp/
-        shwrap("test -d ${cosaDir}/tmp/kola")
+        shwrap("test -d ${outputDir}/kola")
```
jlebon, line 66: Also, shouldn't this be just

```suggestion
        shwrap("test -d ${outputDir}")
```

now?

And similarly below.

dustymabe: not exactly what you suggested, but I reworked it to fix this issue.

jlebon (approved):

PR #96, opened 2022-01-04
fcosKola: don't use `parallel` for single runs

jlebon:
That should make the Blue Ocean view less messy.

Related: https://github.com/coreos/coreos-ci-lib/issues/95

PR #99, opened 2022-01-07
fcosKola: don't use `it` variable

jlebon:
Jenkins is complaining about `it` already existing, which I'm not sure
why. I don't see it existing elsewhere, and the overall pattern works
fine when manually testing in the Jenkins script console. Let's just try
instead using the `k, v` pattern.

Fixes: #96


dustymabe (approved):
LGTM

PR #100, opened 2022-01-07
fcosKola: make archived artifact filenames unique

dustymabe:
Previously if a Jenkins run executed multiple fcosKola calls it would
run multiple times but there would only be one archived kola.tar.xz at
the end of the run because each run overwrote the previous one. Generate
some unique characters to add to the filename of each uploaded file. Also
allow the caller to pass in a marker that will be added to the uploaded
filenames to make them more identifiable.

### DIFF
--- a/vars/fcosKola.groovy
+++ b/vars/fcosKola.groovy
@@ -9,6 +9,7 @@
 //    platformArgs:      string   -- platform-specific kola args (e.g. '-p aws --aws-ami ...`)
 //    extraArgs:         string   -- additional kola args for `kola run` (e.g. `ext.*`)
 //    disableRerun:      boolean  -- disable reruns of failed tests
+//    marker:            string   -- some identifying text to add to uploaded artifact filenames
 def call(params = [:]) {
     def cosaDir = utils.getCosaDir(params)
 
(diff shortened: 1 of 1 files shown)


gursewak1997 (approved):

vars/fcosKola.groovy
```diff
@@ -24,8 +26,8 @@ def call(params = [:]) {
         arch = "--arch=${arch}"
     }
 
-    // Create a unique output directory for this run of fcosKola
-    def outputDir = shwrapCapture("mktemp -d ${cosaDir}/tmp/fcosKola-XXXXX")
```
jlebon, line 28: Hmm, though then if the caller does parallel calls without passing markers, we'll be back to clobbering results, no? (Not only the published artifacts, but we'd also be outputting artifacts which will be an amalgamation of multiple runs.)

Maybe simplest is to default the marker to some generated value? E.g. using `uuidgen | cut -f1 -d-`. At least from the log output they could still match up with parallel branches.

dustymabe: Made a new upload and updated the commit title/message (along with title and description of this PR) to be more appropriate.

PR #101, opened 2022-01-14
fcosKola: make project test name consistent

jlebon:
Right now the kola external tests specific to a project have different
names on each run because it's affected by the workspace directory name,
which Jenkins names something like e.g. hub-ci_coreos_rpm-ostree_PR-3327
which isn't very friendly.

This prevents us from having sane globs and makes output messier.

Let's just get the project name from the git URL and pass a symlink
named as such to kola.


dustymabe (approved):
LGTM

gursewak1997 (approved):

PR #103, opened 2022-02-03
gangplank: print message if command finished cleanly

jlebon:
We're suspecting that the gangplank command may be exiting 0 even when
test failures occurred. Add an echo after to check this: if it's printed
then it must mean that `shwrap` didn't throw an exception.


dustymabe (approved):
LGTM

PR #104, opened 2022-03-04
pod: use `coreos-ci-sa` serviceAccount when using `runAsUser`

jlebon:
That service account has `RunAsAny` caps.

This is a bit hacky. Ideally, this shared lib wouldn't hardcode things
specific to either CoreOS CI or the pipeline here and instead we'd store
this config in https://github.com/coreos/coreos-ci. This works though
because the FCOS pipeline doesn't use `runAsUser` at all.


dustymabe (approved):
LGTM

PR #105, opened 2022-03-04
pod: use serviceAccountName instead of serviceAccount

jlebon:
The latter is deprecated. It should still work at the k8s level, but
it's possible there's something going on with the Jenkins kubernetes
plugin not handling this correctly.


dustymabe (approved):
LGTM

PR #106, opened 2022-03-04
pod: set serviceAccount in podTemplate instead of YAML

jlebon:
Setting the `serviceAccountName` field in the pod YAML itself is not
working with the kubernetes plugin. I noticed that `podTemplate` also
supports specifying the `serviceAccount` directly. It's possible that
that key is used and any service account set in the pod YAML directly is
always overridden.

Let's try using that instead.


dustymabe (approved):
LGTM

PR #107, opened 2022-03-09
Move from Quay.io to app.ci

jlebon:
The latter mirrors to the former, but it's a periodic that runs every
hour. So let's just use the images directly from app.ci for faster
iteration.


ravanelli (approved):
/LGTM

PR #108, opened 2022-03-21
buildImage: add support for setting memory and cpu

jlebon:
In cosa CI, we're hitting:

> runtime: failed to create new OS thread

I think this is another instance of non-Kubernetes-aware multiprocessing
like in https://github.com/coreos/coreos-ci/issues/23.

Let's expose the `resources` knob for building images like we already do
for `pod`. This will allow us in cosa to request a specific amount, and
then asking golang to respect it.


saqibali-2k (approved):
lgtm!

PR #110, opened 2022-05-25
Add podman functions to allow podman remote and manifest creation

ravanelli:


ravanelli:
I will add /hold for now until I can check the rootless issue with Dusty

vars/podman.groovy
```diff
@@ -0,0 +1,85 @@
...
+
+    withCredentials([
+        string(credentialsId: "fcos-${arch}-builder-host-string",
+               variable: 'REMOTEHOST'),
+        string(credentialsId: "fcos-${arch}-builder-uid-string",
+               variable: 'REMOTEUID'),
+        sshUserPrivateKey(credentialsId: "fcos-${arch}-builder-sshkey-key",
+                          usernameVariable: 'REMOTEUSER',
+                          keyFileVariable: 'CONTAINER_SSHKEY')
+    ]) {
```
jlebon, line 19: I don't think we should hardcode credential names in this shared library. Some options:
- remove `withCredentials` here: i.e. have the caller wrap calls to `buildRemote` with `withCredentials` itself
- make the credential names parameters: e.g. have a `host-credential` and a `uid-credential`.

vars/podman.groovy
```diff
@@ -0,0 +1,85 @@
...
+// Available parameters:
+//    tag:               string -- Image tag
+//    repo:              string -- Quay repository
+def createManifest(params = [:]) {
+    repo = params['repo']
+    tag = params['tag'] 
+
+    shwrap("""
+        podman manifest create $repo:main
+        podman manifest add --features cosa_commit=$tag $repo:main docker://$repo:$tag
```
jlebon, line 43: The `cosa_commit` bit here looks like abstraction leak.

vars/podman.groovy
```diff
@@ -0,0 +1,85 @@
...
+// Create and push a manifest list
+// Available parameters:
+//    tag:               string -- Image tag
+//    repo:              string -- Quay repository
+def createManifest(params = [:]) {
+    repo = params['repo']
+    tag = params['tag'] 
+
+    shwrap("""
+        podman manifest create $repo:main
```
jlebon, line 42: Should `main` be parametrized also?

vars/utils.groovy
```diff
@@ -21,6 +21,19 @@ def getCosaDir(params = [:]) {
...
 }
 
+// Clone a git repository in the following dir:
+//    dir:               string -- Directory
+//    git_ref:           string -- Git branch
+//    git_uri:           string -- Git URL
+def gitClone(params = [:]) {
+    def path = params.get('dir', "./src");
+    def branch = params.get('git_uri', "main");
+    def repo = params.get('git_ref', "https://github.com/coreos/coreos-assembler.git");
```
jlebon, line 31: Let's drop this default and let it error out if it's unspecified.

vars/podman.groovy
```diff
@@ -0,0 +1,85 @@
...
+    """)
+} 
+// Run podman remote build and push
+// Available parameters:
+//    arch:              string -- Build architecture
+//    dir:               string -- Directory for source
+//    repo:              string -- Quay repository
+
+def runMultiarchBuild(params = [:]) {
+    def repo = params.get('repo', "https://github.com/coreos/coreos-assembler.git");
```
jlebon, line 60: Let's drop this default and let it error out if it's unspecified.

vars/podman.groovy
```diff
@@ -0,0 +1,85 @@
...
+def cleanImages(params = [:]) {
+    shwrap("""
+        podman rmi -a -f
+    """)
+} 
+// Run podman remote build and push
+// Available parameters:
+//    arch:              string -- Build architecture
+//    dir:               string -- Directory for source
+//    repo:              string -- Quay repository
```
jlebon, line 57: This looks like it's used both as the git repo and the registry repo. Was this supposed to be two separate parameters instead?

dustymabe:
The only part that is special here is the part that uses "withCredentials". I think we should maybe consider making the other pieces be part of COSA itself? `cosa push-manifest` or so?

We should make a generic wrapper for podman remote defined here that we can use in other places:

```
withPodmanRemoteContext("""
    podman --remote build etc..
""")
```

i.e. we'd be creating `withPodmanRemoteContext()` here and then calling that elsewhere. @jlebon - do you know if this is easy or possible?

jlebon:
> The only part that is special here is the part that uses "withCredentials". I think we should maybe consider making the other pieces be part of COSA itself? `cosa push-manifest` or so?
> 
> We should make a generic wrapper for podman remote defined here that we can use in other places:
> 
> ```
> withPodmanRemoteContext("""
>     podman --remote build etc..
> """)
> ```
> 
> i.e. we'd be creating `withPodmanRemoteContext()` here and then calling that elsewhere. @jlebon - do you know if this is easy or possible?

Yup, that's possible. It's a custom step, so it would be similar to things like `cosaPod` or e.g. `shwrap`: https://github.com/coreos/coreos-ci-lib/blob/main/vars/shwrap.groovy.

I think we'd want it to take the credential IDs as parameters.

vars/podman.groovy
```diff
@@ -0,0 +1,27 @@
+//Run podman remote build and push
+// Available parameters:
+//    cmds                     string -- Commands to be executed
```
jlebon, line 3: Instead of passing shell commands as a parameter, we could make this a custom step and let the user do whatever they want within. See e.g. how [`cosaPod`](https://github.com/coreos/coreos-ci-lib/blob/main/vars/cosaPod.groovy) works: the function is named `call` and it takes a `Closure` which we call using `body()`.

vars/utils.groovy
```diff
@@ -21,6 +21,19 @@ def getCosaDir(params = [:]) {
...
 
+// Clone a git repository in the following dir:
+//    dir:               string -- Directory
+//    git_ref:           string -- Git branch
+//    git_uri:           string -- Git URL
+def gitClone(params = [:]) {
+    def path = params.get('dir','');
+    def branch = params.get('git_uri','');
+    def repo = params.get('git_ref', '');
+    dir(path) {
```
jlebon, line 32: Does this work if `path` is the default empty string?

I'm unsure also whether this thin wrapper is worth it. It doesn't seem to provide much value. Given that we usually only clone once per job, just letting the caller use the `git` step or call out to `git clone` seems OK. But fine too if you'd really like to have it!

vars/podman.groovy
dustymabe: Yep. That's what I was thinking.

vars/withPodmanRemoteContext.groovy
```diff
@@ -0,0 +1,25 @@
+//Run podman remote build and push
+// Available parameters:
+//    remotehost:              string -- OCP secret name containing the remote host
+//    remoteuid:               string -- OCP secret name containing the remote user uid
+//    sshkey:                  string -- OCP secret name containing the sshkey
+def call(params = [:], Closure body) {
+
```
jlebon, line 7: Minor: extraneous blank line

vars/withPodmanRemoteContext.groovy
```diff
@@ -0,0 +1,25 @@
+//Run podman remote build and push
```
jlebon, line 1: This needs updating

vars/withPodmanRemoteContext.groovy
```diff
@@ -0,0 +1,25 @@
...
+// Available parameters:
+//    remotehost:              string -- OCP secret name containing the remote host
+//    remoteuid:               string -- OCP secret name containing the remote user uid
+//    sshkey:                  string -- OCP secret name containing the sshkey
```
jlebon, line 5: Some minor tweaks, but also parameter renames for consistency with other functions in this shared lib:

```suggestion
//    remoteHost:              string -- OCP secret name containing the remote host
//    remoteUid:               string -- OCP secret name containing the remote user UID
//    sshKey:                  string -- OCP secret name containing the SSH key
```

jlebon (commented):
@dustymabe You suggested `withPodmanRemoteContext`. WDYT about `withPodmanRemote` instead? It makes it less wordy, but also the `with` naming already implies a context, so it seems redundant.

dustymabe:
> @dustymabe You suggested `withPodmanRemoteContext`. WDYT about `withPodmanRemote` instead?

Yeah I'm OK with that.

vars/withPodmanRemoteContext.groovy
jlebon: Good point. The fact that it's likely backed by a k8s secret is an implementation detail hidden from coreos-ci-lib. The most correct would be to say "Credential ID for (remote host|remote user UID|SSH key)", since that's how we use it.

vars/withPodmanRemote.groovy
```diff
@@ -0,0 +1,25 @@
+//Run podman remote context using the
+//environment variables CONTAINER_SSHKEY and CONTAINER_HOST
```
jlebon, line 2: Minor/optional: missing space after `//`.

jlebon (approved):
Thanks, LGTM. Some comments, but nothing blocking if you'd like to merge as is.

vars/withPodmanRemote.groovy
```diff
@@ -0,0 +1,25 @@
...
+// Available parameters:
+//    remoteHost:    string -- Credential ID containing the remote host
+//    remoteUid:     string -- Credential ID containing the remote user UID
+//    sshKey:        string -- Credential ID containing the private SSH key
+def call(params = [:], Closure body) {
+    withCredentials([
+        string(credentialsId: params['remoteHost'],
+               variable: 'REMOTEHOST'),
+        string(credentialsId: params['remoteUid'],
+               variable: 'REMOTEUID'),
```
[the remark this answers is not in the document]
dustymabe: Yeah. It's easier if we leave it like it is for now.

Right now we've been doing everything through the podman socket (no extra ssh commands) and the only way to build up the path to the socket is with the UID.

dustymabe (approved):

dustymabe:
LGTM

jlebon (approved):
Thanks!

PR #111, opened 2022-07-27
Rename files for accuracy

ravanelli:
- The coreos-ci-lib will also used in RHCOS, let's
remove the FCOS names mention to have a better accuracy
on it


vars/cosBuild.groovy
```diff
@@ -12,7 +12,7 @@
 //    skipInit:       boolean  -- Assume `cosa init` has already been run
 //    skipKola:       boolean  -- Do not automatically run kola on resulting build
 def call(params = [:]) {
-    stage("Build FCOS") {
+    stage("Build COREOS") {
```
jlebon, line 15: Maybe just "Build"?

jlebon (commented):
For the new names, how about: `cosaBuild`, `kola`, and `kolaTestIso`?

Also, we'll want to keep supporting the old names to give time for codebases to migrate over.

Hmm, it's possible that a symlink would work. Otherwise, it's trivial to have e.g. a tiny `fcosKola` step that just calls `kola`.

vars/buildPod.groovy
```diff
@@ -1,4 +1,4 @@
-// buildPod is a thin wrapper around pod() which fills in the FCOS buildroot images.
+// buildPod is a thin wrapper around pod() which fills in the COREOS buildroot images.
```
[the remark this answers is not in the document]
ravanelli: Would only `kola` cause some confusion? Or issues? I was thinking about replacing all the code we have around. The two place we have it is in the `fedora-pipeline` and `cosa` I suppose.

vars/kola.groovy
```diff
@@ -30,8 +30,8 @@ def call(params = [:]) {
     // Prevents multiple runs overwriting same filename in archiveArtifacts
     def token = shwrapCapture("uuidgen | cut -f1 -d-")
 
-    // Create a unique output directory for this run of fcosKola
-    def outputDir = shwrapCapture("mktemp -d ${cosaDir}/tmp/fcosKola-XXXXX")
+    // Create an unique output directory for this run of kola
```
dustymabe, line 33: this should stay "a unique" right?

dustymabe: since unique starts with a Y sound

dustymabe:
`cosKolaTestIso:` in the second commit title seems wrong.

vars/kola.groovy
dustymabe: cc @jlebon since you made the suggestion. I have a feeling I'm about to learn something :)

ravanelli: You can blame me @dustymabe =) fixing it

dustymabe:
Changes seem reasonable. I think we still need to handle migration, though. As jlebon mentioned.

vars/kolaTestIso.groovy
```diff
@@ -43,9 +43,9 @@ def call(params = [:]) {
         testIsoRuns1["metal4k"] = {
             try {
                 if (params['scenarios4k']) {
-                    shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k ${extraArgs4k} --scenarios ${scenarios4k} --output-dir tmp/kola-testiso-metal4k")
+                    shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k --qemu-multipath ${extraArgs4k} --scenarios ${scenarios4k} --output-dir tmp/kola-testiso-metal4k")
                 } else {
-                    shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k ${extraArgs4k} --output-dir tmp/kola-testiso-metal4k")
+                    shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k --qemu-multipath ${extraArgs4k} --output-dir tmp/kola-testiso-metal4k")
```
jlebon, line 48: What's the background for this change? We're already testing multipath lower down. If we want to add coverage for multipath *and* metal4k together, then let's add a new invocation in the block below instead of modifying this one. That way it also gets gated by the `skipMultipath` variable.

Otherwise, we're not getting any vanilla metal4k coverage.

vars/buildPod.groovy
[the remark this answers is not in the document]
jlebon: > Would only kola cause some confusion? Or issues?

I see what you mean, though I think it should be clear for context when `kola` is used as a Groovy identifier vs in a shell. I don't think it should cause any issues.

> I was thinking about replacing all the code we have around. The two place we have it is in the fedora-pipeline and cosa I suppose.

The full list of repos to check is https://github.com/coreos/coreos-ci/blob/1aef511907d5cc3f7b00fa7057e6576ab98bfc66/jobs/seed-github-ci.Jenkinsfile#L4-L19. But to migrate those without breaking CI, we need to ratchet this change in by leaving the old names usable. Once everything is migrated, we can delete them.

vars/kolaTestIso.groovy
ravanelli: All tests for 4k in FCOS and RHCOS that I checked are testing using `--qemu-multipath` as [here](https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/build.Jenkinsfile#L479) . I also assumed FCOS/RHCOS were the only ones using the lib

jlebon: Ahh gotcha. Yeah, ideally we'd have the pipeline use `kolaTestIso` in the future to share this.

vars/fcosBuild.groovy
```diff
@@ -1,60 +0,0 @@
-// Build FCOS, possibly with modifications.
```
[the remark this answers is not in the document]
ravanelli: Yeah, tested it. It seems to work ;)

vars/kolaTestIso.groovy
```diff
@@ -52,6 +53,20 @@ def call(params = [:]) {
...
+                        shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k --qemu-multipath ${extraArgs4k} --scenarios ${scenarios4k} --output-dir tmp/kola-testiso-metal4k")
+                    } else {
+                        shwrap("cd ${cosaDir} &&  kola testiso -S --qemu-native-4k --qemu-multipath ${extraArgs4k} --output-dir tmp/kola-testiso-metal4k")
+                    }
+                } finally {
+                    shwrap("cd ${cosaDir} && tar -cf - tmp/kola-testiso-multipath4k/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-multipath4k.tar.xz")
+                    archiveArtifacts allowEmptyArchive: true, artifacts: 'kola-testiso-multipath4k.tar.xz'
+                }
+            }
+        }
```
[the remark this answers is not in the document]
ravanelli: ummm, I added it out mostly because the tests I saw the multipath test was done only in 4k and not in the other tests, but I'm not so sure if it  makes sense to test it only in 4k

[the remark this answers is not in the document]
jlebon: OK gotcha. There's a combinatorial matrix issue here with UEFI, multipath, and metal4k (though that one implies UEFI). I think let's just drop that commit for now and we can revisit. When we move the FCOS pipeline to use `kolaTestIso()`, we'll still get multipath coverage but on metal, which is a more common scenario than metal4k anyway.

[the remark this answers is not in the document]
ravanelli: a ok. Dropping it

jlebon (approved):
Diff'ed this manually locally because GitHub doesn't understand that the files were copied and slightly modified. I guess one way to do this would've been to split it into 2 commits: one that does pure file renames (`git mv`), and then another that adds the compat symlinks and does the few tweaks in the copied files. Feel free to do that if you'd like but OK with it as is.

```diff
diff --git a/vars/fcosBuild.groovy b/vars/cosaBuild.groovy
similarity index 96%
copy from vars/fcosBuild.groovy
copy to vars/cosaBuild.groovy
index 30996e2..5f481ae 100644
--- a/vars/fcosBuild.groovy
+++ b/vars/cosaBuild.groovy
@@ -1,4 +1,4 @@
-// Build FCOS, possibly with modifications.
+// Build CoreOS, possibly with modifications.
 // Available parameters:
 //    cosaDir:        string   -- Cosa working directory
 //    extraArgs:      string   -- Extra arguments to pass to `cosa build`
@@ -12,7 +12,7 @@
 //    skipInit:       boolean  -- Assume `cosa init` has already been run
 //    skipKola:       boolean  -- Do not automatically run kola on resulting build
 def call(params = [:]) {
-    stage("Build FCOS") {
+    stage("Build") {
         def cosaDir = utils.getCosaDir(params)
         def extraFetchArgs = params.get('extraFetchArgs', "");
         def extraArgs = params.get('extraArgs', "");
@@ -54,7 +54,7 @@ def call(params = [:]) {
     }

     if (!params['skipKola']) {
-        fcosKola()
+        kola()
     }
 }

diff --git a/vars/fcosKola.groovy b/vars/kola.groovy
similarity index 96%
copy from vars/fcosKola.groovy
copy to vars/kola.groovy
index 437fc94..3b167d2 100644
--- a/vars/fcosKola.groovy
+++ b/vars/kola.groovy
@@ -30,8 +30,8 @@ def call(params = [:]) {
     // Prevents multiple runs overwriting same filename in archiveArtifacts
     def token = shwrapCapture("uuidgen | cut -f1 -d-")

-    // Create a unique output directory for this run of fcosKola
-    def outputDir = shwrapCapture("mktemp -d ${cosaDir}/tmp/fcosKola-XXXXX")
+    // Create a unique output directory for this run of kola
+    def outputDir = shwrapCapture("mktemp -d ${cosaDir}/tmp/kola-XXXXX")

     // This is a bit obscure; what we're doing here is building a map of "name"
     // to "closure" which `parallel` will run in parallel. That way, we can
diff --git a/vars/fcosKolaTestIso.groovy b/vars/kolaTestIso.groovy
similarity index 100%
copy from vars/fcosKolaTestIso.groovy
copy to vars/kolaTestIso.groovy
diff --git a/vars/utils.groovy b/vars/utils.groovy
index c57314d..dc6f4e9 100644
--- a/vars/utils.groovy
+++ b/vars/utils.groovy
@@ -17,7 +17,7 @@ def getCosaDir(params = [:]) {
     if (params['cosaDir']) {
         return params['cosaDir']
     } else {
-        return "/srv/fcos"
+        return "/srv/coreos"
     }
 }

```

LGTM!

jlebon:
Can you rebase it?

jlebon:
> I guess one way to do this would've been to split it into 2 commits: one that does pure file renames (`git mv`), and then another that adds the compat symlinks and does the few tweaks in the copied files.

Just for general interest, this is how I would do this:

```command
$ git mv vars/fcosBuild.groovy vars/cosaBuild.groovy # and similarly for other renames
$ git commit -am 'tree: drop FCOS name from custom steps'
$ vim vars/cosaBuild.groovy vars/kola.groovy vars/utils.groovy # do the in-file modifications
$ git commit -am 'tree: drop FCOS wording from various places'
```

(But again, I'm cool without it too! A local `git diff` correctly detects the renames so it'd be just to make GitHub happy.)

dustymabe:
I rebased this PR on top of latest `main` branch.

PR #112, opened 2022-07-27
pod: set both the resource CPU request and limit

jlebon:
Right now, we're not setting any CPU limit in the pods we schedule. This
means our workloads runs without constraints and can hog the hosts
they're running on if there are no namespace-wide default resource
limits.

Unlike with memory requests/limits, limits are enforced at the CPU
scheduling level, so IIUC it's not possible to be evicted due to
excessive CPU usage.

So let's start defaulting to a reasonable not too large value and ensure
we set both the CPU requests and limits. This also allows cgroups-aware
applications to derive the correct number of "equivalent CPUs" to use
for multi-threaded/multi-processing steps.

### DIFF
--- a/vars/pod.groovy
+++ b/vars/pod.groovy
@@ -33,12 +33,20 @@ def call(params = [:], Closure body) {
     if (params['memory']) {
         podObj['spec']['containers'][0]['resources']['requests']['memory'] = params['memory'].toString()
     }
-    if (params['cpu']) {
-        podObj['spec']['containers'][0]['resources']['requests']['cpu'] = params['cpu'].toString()
-        // Also propagate CPU count to NCPUS, because it can be hard in a Kubernetes environment
-        // to determine how much CPU one should really use.
-        podObj['spec']['containers'][0]['env'] += ['name': 'NCPUS', 'value': params['cpu'].toString()]
+    // just default to 2 cpu-equivalent if unspecified
+    if (params['cpu'] == null) {
+        params['cpu'] = "2"
     }
+
+    // note we set *both* the request and limit here; the limit is what
+    // actually affects cgroups scheduling
+    podObj['spec']['containers'][0]['resources']['requests']['cpu'] = params['cpu'].toString()
+    podObj['spec']['containers'][0]['resources']['limits']['cpu'] = params['cpu'].toString()
+
+    // Also propagate CPU count to NCPUS, because it can be hard in a Kubernetes environment
+    // to determine how much CPU one should really use.
+    podObj['spec']['containers'][0]['env'] += ['name': 'NCPUS', 'value': params['cpu'].toString()]
+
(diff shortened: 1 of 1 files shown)


vars/pod.groovy
```diff
@@ -33,12 +33,20 @@ def call(params = [:], Closure body) {
...
-        podObj['spec']['containers'][0]['env'] += ['name': 'NCPUS', 'value': params['cpu'].toString()]
+    // just default to 2 cpu-equivalent if unspecified
+    if (params['cpu'] == null) {
+        params['cpu'] = "2"
     }
+
+    // note we set *both* the request and limit here; the limit is what
+    // actually affects cgroups scheduling
+    podObj['spec']['containers'][0]['resources']['requests']['cpu'] = params['cpu'].toString()
+    podObj['spec']['containers'][0]['resources']['limit']['cpu'] = params['cpu'].toString()
```
bgilbert, line 44:
```suggestion
    podObj['spec']['containers'][0]['resources']['limits']['cpu'] = params['cpu'].toString()
```

ravanelli (commented):
lgtm

ravanelli (approved):

dustymabe:
hmm. I kind of liked the fact that if there were spare resources available we could "go faster". I guess there's no difference between "use resources if they are available" and "prevent other people from using resources we are currently using".

i.e. it would be nice if CPUs are idling for us to use them, but if other pods got scheduled on the node to be able to yield that extra CPU back to them.

vars/pod.groovy
jlebon: Nice catch!

bgilbert (approved):
LGTM

jlebon:
> hmm. I kind of liked the fact that if there were spare resources available we could "go faster". I guess there's no difference between "use resources if they are available" and "prevent other people from using resources we are currently using".
> 
> i.e. it would be nice if CPUs are idling for us to use them, but if other pods got scheduled on the node to be able to yield that extra CPU back to them.

Yeah, this is an intricate topic, and I'm definitely not an SME. But the base problem as I understand it is that there's no reliable way to know how much CPU we should use when there is only a soft limit (e.g. Kubernetes' `request`, which gets translated into `cpu.shares`). It feels wrong to have e.g. a compressor think it can use 64 threads. And in theory being able to use idle CPU sounds nice, but in practice this makes us noisy neighbours which can affect performance (esp. latency) in colocated workloads.

Being able to declare upfront how much CPU we need also means we'll be nicer to others and ourselves (since we schedule multiple pods across multiple namespaces and jobs) and could actually improve CI and pipeline reliability. It might lead to realizing we're not reserving enough CPU, in which case we should feel free in e.g. the pipeline to request more as necessary.

dustymabe (approved):
LGTM

PR #113, opened 2022-08-04
fcosKola: parallelize kola to the number of CPUs

bgilbert:
As of 7b552e1134, we're enforcing CPU resource limits on our pods.  The current defaults allocate 2 CPUs and then use them to run 8 parallel kola tests and a kola upgrade test.  Reduce oversubscription by having kola parallelize to the number of available CPUs by default.  There's still some oversubscription because we're running the kola upgrade test in parallel with that, but that currently only needs 1 CPU.


jlebon (approved):
Nice, thanks!

PR #114, opened 2022-08-08
cosaPod: point back to Quay.io for cosa

jlebon:
The canonical location is Quay.io again so let's use that so we don't
have to wait for app.ci to mirror it.


dustymabe (approved):
LGTM

gursewak1997 (approved):

PR #117, opened 2022-10-11
Add `tryWithCredentials` custom step

jlebon:
This new step can be used just like `withCredentials`, except that if the credentials don't exist, we'll gracefully return.

Upstreamed from the FCOS pipeline.


ravanelli (approved):

dustymabe (approved):
LGTM - obviously need a corresponding PR to the pipeline?

jlebon:
> LGTM - obviously need a corresponding PR to the pipeline?

Yes, but it doesn't need to be done in lockstep... I don't think.

jlebon:
https://github.com/coreos/fedora-coreos-pipeline/pull/689

PR #118, opened 2022-10-12
kola: run provisioning tests separately

jlebon:
This effectively upstreams
https://github.com/coreos/fedora-coreos-pipeline/commit/c3af4ad57ab4.

The same reasons apply, but we're doing it now here too because it'll help with not hitting against our memory limits.

### DIFF
--- a/vars/kola.groovy
+++ b/vars/kola.groovy
@@ -68,10 +68,16 @@ def call(params = [:]) {
             if (!params['skipBasicScenarios']) {
                 shwrap("cd ${cosaDir} && cosa kola run ${rerun} --output-dir=${outputDir}/kola --basic-qemu-scenarios")
             }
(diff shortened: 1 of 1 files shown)


jlebon:
I bet we could revert this (and the pipeline version) once we have https://github.com/coreos/coreos-assembler/issues/3118. Because that'll naturally lower the concurrency on these tests.

Though if it still causes issues, we should lower this into kola in the form of a e.g. `heavyweight` tag or something and then kola knows not to run more than one `heavyweight` test at a time.

dustymabe (approved):
LGTM

cgwalters:
Ah, nice, thanks for digging into this.  I think https://github.com/coreos/coreos-assembler/issues/3118 is the best fix, but even short term, this also touches on https://github.com/coreos/coreos-assembler/issues/3008 and the need to be able to implement logic like this that's decoupled from Jenkins.

PR #119, opened 2022-10-12
utils: add substituteStr

jlebon:
This will be used by the pipeline to process some of the templated fields in `config.yaml`. Putting it here allows us to work around the fact that the `SimpleTemplateEngine` API would normally have to be allowed by an administrator first.

### DIFF
--- a/vars/utils.groovy
+++ b/vars/utils.groovy
@@ -25,3 +25,12 @@ def getCosaDir(params = [:]) {
 def pathExists(path) {
     return shwrapRc("test -e ${path}") == 0
 }
+
+// Thin wrapper around SimpleTemplateEngine().
+//     tmpl_str    string  -- templated string
+//     bindings    map     -- variables to fill in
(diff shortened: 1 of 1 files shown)


vars/utils.groovy
```diff
@@ -25,3 +25,12 @@ def getCosaDir(params = [:]) {
 def pathExists(path) {
     return shwrapRc("test -e ${path}") == 0
 }
+
+// Thin wrapper around SimpleTemplateEngine().
+//     tmpl_str    string  -- templated string
+//     bindings    map     -- variables to fill in
```
dustymabe, line 31: I wonder if we should put some of the context from the commit message about why it was placed in this library in the comment here.

dustymabe (approved):
LGTM

vars/utils.groovy
jlebon: No strong opinion. Even without the API restriction thing, I think it makes sense to have it here to share more easily.

PR #121, opened 2022-10-13
vars/utils: add compat symlink from `/srv/fcos` to `/srv/coreos`

jlebon:
Commit 3719f69 ("tree: drop FCOS wording from various places") broke multiple upstream CI jobs which still reference the old location. We need to temporarily support both locations as we migrate them.

Add a compat symlink.


dustymabe (approved):
LGTM

PR #124, opened 2022-10-17
{kola,kolaTestIso}: workaround issue with accessDenied from dir()

dustymabe:
When using `dir()` in the kola and kolaTestIso jobs the cd is happening on the controller (not in the COSA pod) so if the directory isn't under the env.WORKSPACE directory then we get a permission denied error.

Let's go back to using a `cd ${cosaDir}` for now to workaround this while we find better solutions. This `cd ${cosaDir}` will have no effect on multi-arch where we'll be operating in a remote session anyway.

### DIFF
--- a/vars/kola.groovy
+++ b/vars/kola.groovy
@@ -31,7 +31,7 @@ def call(params = [:]) {
     def token = shwrapCapture("uuidgen | cut -f1 -d-")
 
     // Create a unique output directory for this run of kola
-    def outputDir = shwrapCapture("cosa shell -- mktemp -d ${cosaDir}/tmp/kola-XXXXX")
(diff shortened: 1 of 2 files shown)


jlebon (approved):
LGTM! As discussed OOB, we will circle back on this and rework things to make it cleaner.

> When using dir() in the kola and kolaTestIso jobs the cd is happening on the controller (not in the COSA pod).

To clarify, that was my first theory but in testing, the fact that it does work if it's a subdir of the workspace seems to indicate it's more subtle than that.

dustymabe:
> To clarify, that was my first theory but in testing, the fact that it does work if it's a subdir of the workspace seems to indicate it's more subtle than that.

updated the commit message to drop that possibly inaccurate detail.

jlebon:
> LGTM! As discussed OOB, we will circle back on this and rework things to make it cleaner.

Draft in https://github.com/coreos/coreos-ci-lib/pull/125.

PR #125, opened 2022-10-17
tree: drop support for `cosaDir`

jlebon:
Support for the cosa directory being different from the working
directory makes implementation details messier.

Drop support for it.

Callers can still have their cosa directories be different from the
Jenkins workspace by first wrapping calls to the cosa APIs here in a
call to `dir(cosaDir)`:

```
dir(cosaDir) {
        cosaBuild(skipKola: true)
        kola()
        kolaTestIso(skipMetal4k: true)
}
```

The cosa directory currently needs to be a subdirectory of the workspace
(see https://github.com/coreos/coreos-ci-lib/pull/124).


jlebon:
This requires migrating all users away from `cosaDir` and the other removed functionality. This should be straightforward for upstreams which have already adapted the model of building in `buildPod`, stashing the build, then building FCOS in a `cosaPod`.

dustymabe (approved):
LGTM

PR #128, opened 2022-10-20
utils: add runParallel() function

dustymabe:
This function will behave the same as jenkins parallel but will only run `max` number of runs in parallel at once.


jlebon (approved):
Cool!

PR #130, opened 2022-10-25
kola/kolaTestIso: fix tar path

jlebon:
If we pass the full directory name to `tar -c`, it'll want to recreate the whole structure on extracting. Instead, use `-C` so we only pass the final directory to `tar`. That'll ensure we still create at least one directory on extraction instead of filling up the working directory. We used to do this, but I think it got lost in the recent enhancements on these steps.


dustymabe (approved):
LGTM

PR #131, opened 2022-10-28
pod: add support for `env` parameter

jlebon:
This allows modifying the container's environment variables. This is generally useful, but the immediate use case for it is to add support for a temporary hack knob in the pipeline to easily disable all SecureBoot-related tests, such as when delivering a hotfix with an unsigned kernel.


jlebon:
Actually, I don't think we'll need this.

PR #133, opened 2022-10-28
kola/kolaTestIso: support skipSecureBoot

jlebon:
This will be used by the pipeline when SecureBoot tests should be skipped (e.g. when testing an unsigned kernel).


dustymabe (approved):
LGTM

dustymabe:
going to go ahead and merge this - seems safe enough and we'll have a few runs to test it.

PR #134, opened 2022-11-07
kolaTestIso: add `miniso-install` to UEFI scenarios

jlebon:
This is tested on BIOS already, but it's useful to test on UEFI too because it implicitly also verifies that we can change kargs in the UEFI `grub.cfg`.


dustymabe:
This LGTM, but let's hold it until this week's releases are out.

One other thing here. With the new pipeline work/migration these tests are quickly becoming "global", which means they need to run on older RHCOS releases (even ones that didn't have miniso support). I think we're going to need to accelerate denylist support for testiso bits (even if it's a hacky short term solution).

PR #137, opened 2022-12-16
Drop gangplank.groovy

jlebon:
AFAIK, we've never used these functions and we've moved away from gangplank for now in favour of `podman remote`. Let's delete it. We can always restore it in the future if needed.


dustymabe (approved):
LGTM

PR #138, opened 2022-12-16
Add warn helper

jlebon:
We started using this convention in a few places in the pipeline. Let's add a helper for it. Make it top-level so it feels as ergonomic as an `error` call.


dustymabe (approved):
LGTM

Are we sure this doesn't clash with some existing builtin?

dustymabe:
Note I think we can also use this in this library itself too where we warn.

jlebon:
Nice, made the one caller in this repo use it.
Yes, I've verified that `warn` isn't an existing function.

PR #140, opened 2022-12-22
utils: add `buildAsync` and `waitBuild`

jlebon:
Add two new APIs: `buildAsync` and `waitBuild`. With `buildAsync`, one can start a new build and keep a reference to it that can later be passed to `waitBuild` to wait for it to complete.

The advantages over the built-in `build(wait: false)` are two-fold:
1. it allows the caller to inspect the started build (e.g. build number and status), and
2. it provides the caller with more flexibility over when to start builds and when to wait for them than what `build(wait: true)` in a parallel branch would allow (which would also make code messier and the BlueOcean view busier)

### DIFF
--- a/vars/utils.groovy
+++ b/vars/utils.groovy
@@ -1,3 +1,6 @@
+import org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper
+import hudson.model.Cause.UpstreamCause;
+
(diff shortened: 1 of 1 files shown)


jlebon:
We no longer have an immediate need for these functions. It could come useful in the future when we want more flexibility around triggering, as which point we could revive it.

PR #141, opened 2023-01-09
buildImage: put tarball in directory itself

jlebon:
Currently, we're putting the tarball in `/tmp`, but if we're running on
the Jenkins controller, that's a semi-permanent shared location. Since
our generated tarball has a predictable name, we additionally run the
risk of racing with other `buildImage` calls on the controller and then
passing the wrong `dir.tar` to `oc start-build`.

Fix this by putting the tarball in the workspace itself. That way, it's
properly lifecycled to the job run, and there's no chance of racing with
other jobs.

### DIFF
--- a/vars/buildImage.groovy
+++ b/vars/buildImage.groovy
@@ -47,14 +47,14 @@ def call(params = [:]) {
                 def jobMap  = job.object()
                 def uid     = jobMap.metadata.get('uid')
 
-                imageName = "coreos-ci-$repo-${UUID}"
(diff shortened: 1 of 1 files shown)


jlebon:
> we additionally run the
risk of racing with other `buildImage` calls on the controller and then
passing the wrong `dir.tar` to `oc start-build`.

This actually happened in https://github.com/coreos/coreos-assembler/pull/3311.

dustymabe (approved):
trivial LGTM.

I don't understand the problem 100% but the changes seem logical enough.

PR #142, opened 2023-01-22
kola: fix allowUpgradeFail by removing id from log check

dustymabe:
If the kola upgrade test fails early enough that logs werene't even created then we'll get a failure like:

```
10:50:22  + cosa shell -- test -d /home/jenkins/agent/workspace/build/tmp/kola-5JOIO/kola-upgrade
10:50:22  error: failed to execute cmd-shell: exit status 1
```

Let's try to remove the log from consideration if that's the case.

### DIFF
--- a/vars/kola.groovy
+++ b/vars/kola.groovy
@@ -120,11 +120,15 @@ def call(params = [:]) {
             // If upgrades are broken `cosa kola --upgrades` might
             // fail to even find the previous image so we wrap this
             // in a try/catch so allowUpgradeFail can work.
+            def id = marker == "" ? "kola-upgrade" : "kola-upgrade-${marker}"
+            ids += id
             try {
-                def id = marker == "" ? "kola-upgrade" : "kola-upgrade-${marker}"
-                ids += id
                 shwrap("cd ${cosaDir} && cosa kola ${rerun} --output-dir=${outputDir}/${id} --upgrades --build=${buildID} ${archArg} ${platformArgs}")
             } catch(e) {
+                // If we didn't even get logs then let's remove them from the list
+                if (shwrapRc("cd ${cosaDir} && cosa shell -- test -d ${outputDir}/${id}") != 0) {
+                    ids.remove(id)
+                }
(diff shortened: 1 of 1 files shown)


vars/kola.groovy
```diff
@@ -125,6 +125,11 @@ def call(params = [:]) {
                 ids += id
                 shwrap("cd ${cosaDir} && cosa kola ${rerun} --output-dir=${outputDir}/${id} --upgrades --build=${buildID} ${archArg} ${platformArgs}")
             } catch(e) {
+                // If we didn't even get logs then let's remove them from the list
+                if (shwrapRc("cd ${cosaDir} && cosa shell -- test -d ${outputDir}/${ids[-1]}") != 0) {
+                    ids.remove(ids.size() - 1)
+                }
+                shwrap("cd ${cosaDir} && cosa shell -- test -d ${outputDir}/${id}")
```
jlebon, line 132: Is this leftover?

vars/kola.groovy
```diff
@@ -125,6 +125,11 @@ def call(params = [:]) {
                 ids += id
                 shwrap("cd ${cosaDir} && cosa kola ${rerun} --output-dir=${outputDir}/${id} --upgrades --build=${buildID} ${archArg} ${platformArgs}")
             } catch(e) {
+                // If we didn't even get logs then let's remove them from the list
+                if (shwrapRc("cd ${cosaDir} && cosa shell -- test -d ${outputDir}/${ids[-1]}") != 0) {
+                    ids.remove(ids.size() - 1)
+                }
```
jlebon, line 131: Testing in the console, passing the element itself to `remove` also works. I think it'd be cleaner than using indices. We can hoist the `def id` up to outside the `try` block, and then here we can just use `${outputDir}/${id}` and `ids.remove(id)`.

vars/kola.groovy
dustymabe: yep. it was extraneous.

jlebon (approved):

PR #143, opened 2023-02-09
kolaTestIso: Update to reflect new kola testiso in coreos-assembler

ravanelli:
- Clean up code to reflect changes done in kola testiso in coreos-assembler

### DIFF
--- a/vars/kolaTestIso.groovy
+++ b/vars/kolaTestIso.groovy
@@ -1,102 +1,38 @@
 // Run kola testiso
 // Available parameters:
-//     arch:               string  -- the target architecture
 //     cosaDir:            string  -- cosa working directory
 //     extraArgs:          string  -- extra arguments to pass to `kola testiso`
-//     extraArgs4k:        string  -- extra arguments to pass to 4k `kola testiso`
-//     extraArgsMultipath: string  -- extra arguments to pass to multipath `kola testiso`
-//     extraArgsUEFI:      string  -- extra arguments to pass to UEFI `kola testiso`
-//     scenarios:          string  -- scenarios to pass to `kola testiso`
-//     scenarios4k:        string  -- scenarios to pass to `kola testiso`
-//     scenariosMultipath: string  -- scenarios to pass to `kola testiso`
-//     scenariosUEFI:      string  -- scenarios to pass to `kola testiso`
-//     skipMetal4k:        boolean -- skip metal4k image
-//     skipMultipath:      boolean -- skip multipath tests
+//     marker:             string  -- some identifying text to add to uploaded artifact filenames
 //     skipUEFI:           boolean -- skip UEFI tests
 //     skipSecureBoot      boolean -- skip secureboot tests
-//     marker:             string  -- some identifying text to add to uploaded artifact filenames
+
 def call(params = [:]) {
-    def arch = params.get('arch', "x86_64");
-    def cosaDir = utils.getCosaDir(params)
+    def cosaDir = utils.getCosaDir(params);
     def extraArgs = params.get('extraArgs', "");
-    def extraArgs4k = params.get('extraArgs4k', "");
-    def extraArgsMultipath = params.get('extraArgsMultipath', "");
-    def extraArgsUEFI = params.get('extraArgsUEFI', "");
-    def scenarios = params.get('scenarios', "");
-    def scenarios4k = params.get('scenarios4k', "");
-    // only one test by default
-    def scenariosMultipath = params.get('scenariosMultipath', "iso-offline-install");
-    // by default, only test that we can boot successfully
-    def scenariosUEFI = params.get('scenariosUEFI', "iso-live-login,iso-as-disk");
     def marker = params.get('marker', "");
 
+
+    if (params['skipUEFI']) {
+        extraArgs += " --denylist-test *.uefi*"
+    }
+    if (params['skipSecureBoot']) {
+        extraArgs += " --denylist-test *.uefi-secure"
+    }
     // Define a unique token to be added to the file name uploads
     // Prevents multiple runs overwriting same filename in archiveArtifacts
     def token = shwrapCapture("uuidgen | cut -f1 -d-")
 
-    // Create a unique output directory for this run of fcosKola
-    def outputDir = shwrapCapture("cd ${cosaDir} && cosa shell -- mktemp -d ${cosaDir}/tmp/kolaTestIso-XXXXX")
+    // Create a unique output directory for this run of kola
+    def outputDir = shwrapCapture("cd ${cosaDir} && cosa shell -- mktemp -d ${cosaDir}/tmp/kola-XXXXX")
 
-    // list of identifiers for each run for log collection
-    def ids = []
+    def id = marker == "" ? "kolatestiso" : "kolatestiso-${marker}"
 
-    // If given a marker then add it to the parallel run stage titles
-    def titleMarker = marker == "" ? "" : "${marker}:"
-
-    def testIsoRuns = [:]
-    testIsoRuns["${titleMarker}kola:metal"] = {
-        def id = marker == "" ? "kola-testiso-metal" : "kola-testiso-metal-${marker}"
-        ids += id
-        def scenariosArg = scenarios == "" ? "" : "--scenarios ${scenarios}"
-        shwrap("cd ${cosaDir} && cosa kola testiso -S ${extraArgs} ${scenariosArg} --output-dir ${outputDir}/${id}")
-    }
-    if (!params['skipMetal4k']) {
-        // metal4k test doesn't work on s390x for now
-        // https://github.com/coreos/fedora-coreos-tracker/issues/1261
-        // and testiso for s390x doesn't support iso installs either
-        if (arch != 's390x') {
-            testIsoRuns["${titleMarker}kola:metal4k"] = {
-                def id = marker == "" ? "kola-testiso-metal4k" : "kola-testiso-metal4k-${marker}"
-                ids += id
-                def scenariosArg = scenarios4k == "" ? "" : "--scenarios ${scenarios4k}"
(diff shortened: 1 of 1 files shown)


vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,9 @@
...
-    }
-
-    try {
-        // Run at most two testiso runs at a time to try not to
-        // exceed 8G of memory usage.
-        utils.runParallel(testIsoRuns, 2)
-    } finally {
-        for (id in ids) {
-            shwrap("cd ${cosaDir} && cosa shell -- tar -C ${outputDir} -c --xz ${id} > ${env.WORKSPACE}/${id}-${token}.tar.xz || :")
-            archiveArtifacts allowEmptyArchive: true, artifacts: "${id}-${token}.tar.xz"
```
jlebon, line 99: We still need archiving of test results, no?

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,9 @@
...
-//     extraArgsUEFI:      string  -- extra arguments to pass to UEFI `kola testiso`
-//     scenarios:          string  -- scenarios to pass to `kola testiso`
-//     scenarios4k:        string  -- scenarios to pass to `kola testiso`
-//     scenariosMultipath: string  -- scenarios to pass to `kola testiso`
-//     scenariosUEFI:      string  -- scenarios to pass to `kola testiso`
-//     skipMetal4k:        boolean -- skip metal4k image
-//     skipMultipath:      boolean -- skip multipath tests
-//     skipUEFI:           boolean -- skip UEFI tests
-//     skipSecureBoot      boolean -- skip secureboot tests
-//     marker:             string  -- some identifying text to add to uploaded artifact filenames
```
jlebon, line 17: We still need these two knobs. `marker` is used by the `bump-lockfile` job which runs `kola testiso` for multiple arches in the same run (so needs a way to name the artifacts differently). `skipSecureBoot` is used by https://github.com/coreos/fedora-coreos-pipeline/blob/85a172f6f4f1774ac105ca1e429f3ec74b29a1e0/docs/config.yaml#L31. Ahh, unless you mean we'll use `--denylist-test` from the pipeline to do this now? Would probably be cleaner to encapsulate that here instead so it's consistent with `kola()` (which also takes a `skipSecureBoot` parameter).

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,9 @@
 // Run kola testiso
 // Available parameters:
-//     arch:               string  -- the target architecture
 //     cosaDir:            string  -- cosa working directory
-//     extraArgs:          string  -- extra arguments to pass to `kola testiso`
```
jlebon, line 5: We should probably keep this for flexibility and parity with `kola()`.

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,9 @@
...
-                    shwrap("cd ${cosaDir} && cosa kola testiso -S --qemu-firmware=uefi-secure ${extraArgsUEFI} --scenarios ${scenariosUEFI} --output-dir ${outputDir}/${id}/secure")
-                }
-            }
-        }
-    }
-
-    try {
-        // Run at most two testiso runs at a time to try not to
-        // exceed 8G of memory usage.
-        utils.runParallel(testIsoRuns, 2)
```
jlebon, line 95: This is an interesting consideration. `kola testiso` doesn't support parallel execution right now, but we've hacked around this here by running it in parallel. We lose that with this work. I think that's OK, but just noting it since it'll make the pipeline a bit slower. Long-term, this will be fixed when we merge into `kola run`. Though shorter-term, we could support something similar to https://github.com/coreos/coreos-assembler/pull/3272 in `kola testiso` to reintroduce some parallelism here. But let's not block on that!

jlebon (commented):
Some comments but overall I'm loving how much we'll get to drain from this file!

dustymabe:
I took a look at this. I have no additional comments to add other than what @jlebon mentioned.

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,26 @@
...
-    } finally {
-        for (id in ids) {
-            shwrap("cd ${cosaDir} && cosa shell -- tar -C ${outputDir} -c --xz ${id} > ${env.WORKSPACE}/${id}-${token}.tar.xz || :")
-            archiveArtifacts allowEmptyArchive: true, artifacts: "${id}-${token}.tar.xz"
-        }
-    }
+    shwrap("cd ${cosaDir} && cosa kola testiso -S ${extraArgs} ${denyList} --output-dir ${outputDir}/${id}")
```
dustymabe, line 23: for the `-S` here can we change it to the long form of `--inst-insecure` and comment why we need that? I have to look it up everytime.

```
// We add --inst-insecure here since in CI and in our build pipeline
// the signatures for the metal images won't have been created yet.
```

dustymabe (approved):
Looking good

dustymabe (approved):

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,28 @@
...
-        utils.runParallel(testIsoRuns, 2)
-    } finally {
-        for (id in ids) {
-            shwrap("cd ${cosaDir} && cosa shell -- tar -C ${outputDir} -c --xz ${id} > ${env.WORKSPACE}/${id}-${token}.tar.xz || :")
-            archiveArtifacts allowEmptyArchive: true, artifacts: "${id}-${token}.tar.xz"
-        }
-    }
+    // We add --inst-insecure here since in CI and in our build pipeline
+    // the signatures for the metal images won't have been created yet.
+    shwrap("cd ${cosaDir} && cosa kola testiso --inst-insecure ${extraArgs} ${denyList} --output-dir ${outputDir}/${id}")
```
jlebon, line 25: Shouldn't `denyList` here be processed a bit so it's (1) a list of strings instead of a string, and (2) converted into multiple `--denylist-test` entries? Otherwise, it's no different than `extraArgs`.

But that said, I think what we want instead is to keep `skipSecureBoot` and convert that into `--denylist-test '*.uefi-secure'` (again, I think matching `kola()` on this is appealing). If the caller really needs `--denylist-test`, they can use `extraArgs`.

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,28 @@
...
+    shwrap("cd ${cosaDir} && cosa shell -- tar -C ${outputDir} -c --xz ${id} > ${env.WORKSPACE}/${id}-${token}.tar.xz || :")
+    archiveArtifacts allowEmptyArchive: true, artifacts: "${id}-${token}.tar.xz"
```
jlebon, line 27: This still needs to be in a `try`/`finally` so that we archive test results when a test failed.

jlebon (approved):
LGTM. Nice work!

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,35 @@
...
-//     extraArgs4k:        string  -- extra arguments to pass to 4k `kola testiso`
-//     extraArgsMultipath: string  -- extra arguments to pass to multipath `kola testiso`
-//     extraArgsUEFI:      string  -- extra arguments to pass to UEFI `kola testiso`
-//     scenarios:          string  -- scenarios to pass to `kola testiso`
-//     scenarios4k:        string  -- scenarios to pass to `kola testiso`
-//     scenariosMultipath: string  -- scenarios to pass to `kola testiso`
-//     scenariosUEFI:      string  -- scenarios to pass to `kola testiso`
-//     skipMetal4k:        boolean -- skip metal4k image
-//     skipMultipath:      boolean -- skip multipath tests
-//     skipUEFI:           boolean -- skip UEFI tests
```
jlebon, line 15: Actually, I think there's one last PR we need for the pipeline. It currently uses `skipUEFI` here: https://github.com/coreos/fedora-coreos-pipeline/blob/fe2d907471cd7c73b01b9eca8c3ba4d4b1b1ab44/jobs/build.Jenkinsfile#L365. We'll need to convert that to something like `extraArgs: "--denylist-test *.uefi*"`?

ravanelli: How about adding `skipUEFI` here too?

jlebon: SGTM! Let's redocument it as an option here?

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,36 @@
...
-    // only one test by default
-    def scenariosMultipath = params.get('scenariosMultipath', "iso-offline-install");
-    // by default, only test that we can boot successfully
-    def scenariosUEFI = params.get('scenariosUEFI', "iso-live-login,iso-as-disk");
     def marker = params.get('marker', "");
+    def denyList = "--denylist-test ";
 
+    if (params['skipSecureBoot']) {
+        denyList += "*.uefi-secure"
+    } else if (params['skipUEFI']) {
```
jlebon, line 16: I think we need to flip this around so that we check `skipUEFI` first and then `skipSecureBoot` second. Or alternatively, don't make it an `else if`, but just a separate if statement.

vars/kolaTestIso.groovy
```diff
@@ -1,102 +1,39 @@
...
-    def scenariosUEFI = params.get('scenariosUEFI', "iso-live-login,iso-as-disk");
     def marker = params.get('marker', "");
+    def denyList = "--denylist-test ";
 
+
+    if (params['skipUEFI']) {
+        denyList += "*.uefi*"
+    }
+    if (params['skipSecureBoot']) {
+        denyList += "*.uefi-secure"
```
jlebon, line 20: If both `skipUEFI` and `skipSecureBoot` are set, this will produce `--denylist-test *.uefi* *.uefi-secure`, which is incorrect. I think we should initialize `denyList` as empty, and then add `--denylist-test ...` each time.

Also could append it to `extraArgs` instead of having a separate variable for it, but cool as is too.

PR #144, opened 2023-02-16
Add `HOME` workaround to other shwrap helpers

jlebon:
We had it in `shwrap`, but not `shwrapCapture` and `shwrapRc`. We're hitting an issue right now where `podman` wants to create `$HOME/.ssh` when using the remote stuff but because we're running unprivileged, we get

```
Error: failed to connect: ssh: handshake failed: mkdir /.ssh: permission denied
```

We're still trying to determine why this is new, but let's fix the inconsistency for now.

### DIFF
--- a/vars/shwrapCapture.groovy
+++ b/vars/shwrapCapture.groovy
@@ -1,6 +1,8 @@
 def call(cmds) {
-    return sh(returnStdout: true, script: """
-        set -euo pipefail
-        ${cmds}
-    """).trim()
+    withEnv(["HOME=${env.WORKSPACE}"]) {
+        return sh(returnStdout: true, script: """
+            set -euo pipefail
+            ${cmds}
+        """).trim()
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
Looks sane to me, though I wonder why we don't globally set `$HOME` in the pod?

jlebon:
> Looks sane to me, though I wonder why we don't globally set `$HOME` in the pod?

Good point, will do that in a follow-up. For now, let's see if that unbreaks the pipeline.

jlebon:
https://github.com/containers/common/pull/1216/files#r1109040318

jlebon:
> > Looks sane to me, though I wonder why we don't globally set `$HOME` in the pod?
> 
> Good point, will do that in a follow-up.

https://github.com/coreos/coreos-ci-lib/pull/145

PR #145, opened 2023-02-16
Set `HOME` at the pod level instead of in shwrap helpers

jlebon:
That way we're sure it's always set.


jlebon:
This was reverted in https://github.com/coreos/coreos-ci-lib/pull/146.

PR #146, opened 2023-03-31
Revert "Set `HOME` at the pod level instead of in shwrap helpers"

jlebon:
This reverts commit 3577892a9cac46964a5f44be8881a9ef7741df8c.

Setting the `HOME` var to the workspace from the pod definition doesn't work because it introduces a chicken-and-egg problem: the workspace isn't yet allocated since the pod isn't running yet.

This isn't a pure revert since I wanted to keep 33d4910 ("Add `umask` workaround to other shwrap helpers"), which came after.

### DIFF
--- a/vars/shwrapRc.groovy
+++ b/vars/shwrapRc.groovy
@@ -1,10 +1,13 @@
 def call(cmds) {
-    // If umask is somehow unset, fix it.
-    return sh(returnStatus: true, script: """
-        set -xeuo pipefail
-        if [ `umask` = 0000 ]; then
-          umask 0022
-        fi
-        ${cmds}
-    """)
+    // default is HOME=/ which normally we don't have access to.
+    // Also if umask is somehow unset, fix it.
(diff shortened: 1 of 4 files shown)


vars/shwrapRc.groovy
```diff
@@ -1,10 +1,13 @@
 def call(cmds) {
-    // If umask is somehow unset, fix it.
+    // default is HOME=/ which normally we don't have access to.
+    // Also if umask is somehow unset, fix it.
+    withEnv(["HOME=${env.WORKSPACE}"]) {
     return sh(returnStatus: true, script: """
```
bgilbert, line 5: Incorrect indent.

bgilbert (approved):
LGTM

dustymabe (approved):

vars/shwrapRc.groovy
jlebon: Good catch, thanks! Kinda asked for it reviewing the patch with whitespace ignored. :)

PR #148, opened 2023-05-05
kola: add support for allowRerunSuccess

dustymabe:
We'll start now to default to passing in --allow-rerun-success tags=needs-internet so that we can get notified less about intermittent network flakes.

If someone wants to disable this they can pass in allowRerunSuccess=false.

If someone wants to override the tags to use for rerun-success they can use allowRerunSuccessArgs. For example, allowRerunSuccessArgs="tags=all".

### DIFF
--- a/vars/kola.groovy
+++ b/vars/kola.groovy
@@ -1,18 +1,20 @@
 // Run kola tests on the latest build in the cosa dir
 // Available parameters:
-//    addExtTests:       []string -- list of test paths to run
-//    allowUpgradeFail   boolean  -- warn instead of fail on upgrade failure
-//    arch:              string   -- the target architecture
-//    cosaDir:           string   -- cosa working directory
-//    parallel:          integer  -- number of tests to run in parallel (default: # CPUs)
-//    skipBasicScenarios boolean  -- skip basic qemu scenarios
-//    skipSecureBoot     boolean  -- skip secureboot tests
-//    skipUpgrade:       boolean  -- skip running `cosa kola --upgrades`
-//    build:             string   -- cosa build ID to target
-//    platformArgs:      string   -- platform-specific kola args (e.g. '-p aws --aws-ami ...`)
-//    extraArgs:         string   -- additional kola args for `kola run` (e.g. `ext.*`)
-//    disableRerun:      boolean  -- disable reruns of failed tests
-//    marker:            string   -- some identifying text to add to uploaded artifact filenames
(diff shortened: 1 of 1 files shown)


vars/kola.groovy
```diff
@@ -21,9 +23,14 @@ def call(params = [:]) {
     def buildID = params.get('build', "latest");
     def arch = params.get('arch', "x86_64");
     def marker = params.get('marker', "");
-    def rerun = "--rerun"
-    if (params.get('disableRerun', false)) {
-        rerun = ""
+    def rerun = ""
+    if (!params.get('disableRerun', false)) {
+        rerun = "--rerun"
+        if (!params.get('allowRerunSuccess', false)) {
```
jlebon, line 29: Shouldn't this be

```suggestion
        if (params.get('allowRerunSuccess', true)) {
```

so we default to on?

vars/kola.groovy
```diff
@@ -1,18 +1,20 @@
...
-//    skipUpgrade:       boolean  -- skip running `cosa kola --upgrades`
-//    build:             string   -- cosa build ID to target
-//    platformArgs:      string   -- platform-specific kola args (e.g. '-p aws --aws-ami ...`)
-//    extraArgs:         string   -- additional kola args for `kola run` (e.g. `ext.*`)
-//    disableRerun:      boolean  -- disable reruns of failed tests
-//    marker:            string   -- some identifying text to add to uploaded artifact filenames
+//    addExtTests:          []string -- list of test paths to run
+//    allowUpgradeFail      boolean  -- warn instead of fail on upgrade failure
+//    allowRerunSuccess     boolean  -- whether or not to allowRerunSuccess
+//    allowRerunSuccessArgs string   -- args to pass to --allow-rerun-success
```
jlebon, line 6: We should mention the default for `allowRerunSuccessArg`. For `allowRerunSuccess`, should we make it e.g. `disallowRerunSuccess` for consistency with other "negative" options we have and so it's obvious what the default it?

vars/kola.groovy
jlebon: Alternatively, I think this is correct as is if we call it `disallowRerunSuccess`.

dustymabe:
pushed up a rework which I think addresses code review comments

vars/kola.groovy
dustymabe: went with `disableRerunSuccess` here since there was another `disable` option (`disableRerun`).

PR #150, opened 2023-06-15
pod: Allow caller to specify name of pod

aaradhak:
Allowing the caller to specify the pod name would make it easier to identify what pods were created for what purpose.

We are planning to use this in the debug pod pipeline job to allow users to identify which debug pods they own.

Ref: https://github.com/coreos/fedora-coreos-pipeline/issues/803

### DIFF
--- a/vars/pod.groovy
+++ b/vars/pod.groovy
@@ -10,6 +10,7 @@
 //   serviceAccount: string
 //   secrets: []string secret name and/or my-secret:my-key:mount-path
 //   configMaps: []string
+//   name : string  - optional. The name of the pod we want to create
 def call(params = [:], Closure body) {
     def podJSON = libraryResource 'com/github/coreos/pod.json'
     def podObj = readJSON text: podJSON
@@ -104,6 +105,10 @@ def call(params = [:], Closure body) {
 
     // XXX: look into converting to a YAML string instead
     def label = "pod-${UUID.randomUUID().toString()}"
+    if (params['name']) {
+    	label = params['name']
+    }
+    
     def podYAML
     node {
         writeYaml(file: "${label}.yaml", data: podObj)


vars/pod.groovy
```diff
@@ -104,6 +106,18 @@ def call(params = [:], Closure body) {
...
+    if (params['name']) {
+        def extractedString = ""
+        def regexPattern = /pod-(\w+)-.+/
+        def matcher = (label =~ regexPattern)
+            if (matcher.find()) {
+                extractedString = matcher.group(1)
+                println("extract string: ${extractedString}")
+            }
+        label = "debug-pod-" + params['name'] + "-" + params['arches'] + "-" + extractedString
+        println("LABEL PRINT: ${label}")
```
dustymabe, line 118: What's the point of this code? If the caller passes in the name then we should just use it?

vars/pod.groovy
```diff
@@ -10,6 +10,8 @@
 //   serviceAccount: string
 //   secrets: []string secret name and/or my-secret:my-key:mount-path
 //   configMaps: []string
+//   name : string
```
dustymabe, line 13:
```suggestion
//   name : string - optional. The name of the pod we want to create
```

vars/pod.groovy
```diff
@@ -10,6 +10,8 @@
...
+//   arches : string
```
dustymabe, line 14: the caller should set the name by just handing this function the string to use. We shouldn't be passing this information in here.

dustymabe (changes requested):

vars/pod.groovy
dustymabe: i.e. this whole block should probably be:
```
if (params['name']) {
    label = params['name']
}
```

aaradhak: yes, if the caller passes in the name from the debug-pod, then this "if statement' gets invoked. The name is right now passed only from the debug-pod job. The string in the label is then tweaked to print the desired pod name in the output.

dustymabe (approved):
LGTM - but let's wait to merge this until we finalize https://github.com/coreos/fedora-coreos-pipeline/pull/844

dustymabe (changes requested):
Actually. Please squash this down to one commit.

dustymabe (approved):
LGTM

PR #155, opened 2023-11-20
job.yaml: fix apiVersion and other tweaks

jlebon:
The API version for templates is now `template.openshift.io/v1`. Possibly `v1` was accepted before but not anymore? Fix it.

While we're here, tweak the template name to make it clearer that the template itself is never created as an object.

Also remove the parameter default value to hard require clients to always specify it.

### DIFF
--- a/resources/com/github/coreos/job.yaml
+++ b/resources/com/github/coreos/job.yaml
@@ -3,9 +3,11 @@
 # here isn't used for anything (other than staying alive for 24h).
 
 kind: Template
-apiVersion: v1
+apiVersion: template.openshift.io/v1
 metadata:
-  name: clean-up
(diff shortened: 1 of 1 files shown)


jlebon:
This *should* fix cosa CI. I think it broke with the recent CoreOS CI update because we're now using a newer `oc` client in the base Jenkins image that no longer accepts `v1` as the template apiVersion?

dustymabe (approved):
LGTM

jlebon:
Had missed another instance of this. Since it's trivial, I just pushed it straight to main: https://github.com/coreos/coreos-ci-lib/commit/b7c3eeba0e5caaab0742e0ac590989ffec6ea15a

PR #158, opened 2024-02-20
cosaBuild: add `srcConfig` and `variant` parameters

jlebon:
I want to start building CentOS Stream CoreOS in CoreOS CI for at least openshift/os and coreos/coreos-assembler.

Let's make it easier to point to a different source config than the default FCOS.


dustymabe (approved):
LGTM

PR #159, opened 2024-06-21
vars/kola: Add support for skipping a list of kola tags

marmijo:
Introduce a kola parameter to allow skipping a list of kola tags. This is particularly useful when adding new streams to the pipeline.

xref: https://github.com/coreos/fedora-coreos-pipeline/pull/1008

See: https://github.com/coreos/fedora-coreos-pipeline/issues/1002

### DIFF
--- a/vars/kola.groovy
+++ b/vars/kola.groovy
@@ -10,6 +10,7 @@
 //    skipBasicScenarios  boolean  -- skip basic qemu scenarios (ignored if cosa has https://github.com/coreos/coreos-assembler/pull/3652)
 //    skipSecureBoot      boolean  -- skip secureboot tests
 //    skipUpgrade:        boolean  -- skip running `cosa kola --upgrades`
+//    skipKolaTags:       []string -- list of kola tags to skip
 //    build:              string   -- cosa build ID to target
 //    platformArgs:       string   -- platform-specific kola args (e.g. '-p aws --aws-ami ...`)
 //    extraArgs:          string   -- additional kola args for `kola run` (e.g. `ext.*`)
@@ -102,6 +103,12 @@ def call(params = [:]) {
             args += " --exttest=${env.WORKSPACE}/${path}"
         }
 
+        def skipKolaTags = params.get('skipKolaTags', [])
+
+        for (tag in skipKolaTags) {
+            args += " --tag=!${tag}"
+        }
+
         if (platformArgs != "" || extraArgs != "") {
             // There are two cases where we land here:
             //   1. The user passed `platformArgs`, which implies we're
@@ -114,7 +121,7 @@ def call(params = [:]) {
             //      do a single run in that case.
             id = marker == "" ? "kola" : "kola-${marker}"
             ids += id
-            runKola(id, 'run', "--parallel=${parallel} ${args} ${extraArgs}")
+            runKola(id, 'run', "${args} ${extraArgs} --parallel=${parallel}")
         } else {
             // Check if the basic tests are registered kola tests or not.
             def availableTests = shwrapCapture("kola list --json | jq -r '.[].Name'").split()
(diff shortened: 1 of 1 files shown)


vars/kola.groovy
```diff
@@ -102,6 +103,13 @@ def call(params = [:]) {
             args += " --exttest=${env.WORKSPACE}/${path}"
         }
 
+        def skipKolaTags = params.get('skipKolaTags', "")
+        def skipKolaTagsArgs = ""
```
jlebon, line 107: How about instead of this new variable, we just keep modifying `args` like we do for other directives?

vars/kola.groovy
```diff
@@ -102,6 +103,13 @@ def call(params = [:]) {
...
+
+        for (tag in skipKolaTags) {
+            skipKolaTagsArgs += "--tag=!${tag} "
```
jlebon, line 110: So then, this would be

```suggestion
            args += " --tag=!${tag}"
```

vars/kola.groovy
```diff
@@ -157,7 +165,7 @@ def call(params = [:]) {
             def id = marker == "" ? "kola-upgrade" : "kola-upgrade-${marker}"
             ids += id
             try {
-                runKola(id, 'run-upgrade',  "--upgrades")
+                runKola(id, 'run-upgrade',  "--upgrades ${skipKolaTagsArgs}")
```
jlebon, line 168: I think let's keep out upgrade tests from this for now. It currently doesn't run for RHCOS and those tests should instead be converted into regular kola tests (see also discussions in https://github.com/coreos/coreos-assembler/issues/3714).

jlebon (commented):
Thanks for working on this!

vars/kola.groovy
marmijo: Will do, I wasnt sure if this would be required for the upgrade tests. I'll remove it.

vars/kola.groovy
marmijo: I had thought about that, but we `args` isnt passed to all of the `kola()` runs. Now that we passing this  to all of them, I'll simplify it and just append to `args` :+1:

vars/kola.groovy
```diff
@@ -140,7 +147,7 @@ def call(params = [:]) {
             // normal run (without reprovision tests because those require a lot of memory)
             id = marker == "" ? "kola" : "kola-${marker}"
             ids += id
-            runKola(id, 'run', "--tag='!reprovision' --parallel=${parallel} ${args}")
+            runKola(id, 'run', "--tag='!reprovision' ${args} --parallel=${parallel}")
```
marmijo, line 150: I moved the position of of `${args}` here so the output command would be cleaner. Instead of 
```text
kola run --tag=!reprovision --parallel=5 --tag=!openshift 
```
it'll look like:
```text
kola run --tag=!reprovision --tag=!openshift --parallel=5 
```

vars/kola.groovy
```diff
@@ -102,6 +103,12 @@ def call(params = [:]) {
             args += " --exttest=${env.WORKSPACE}/${path}"
         }
 
+        def skipKolaTags = params.get('skipKolaTags', "")
```
jlebon, line 106:
```suggestion
        def skipKolaTags = params.get('skipKolaTags', [])
```

vars/kola.groovy
jlebon: Sounds good, though for consistency, let's also do that on line 124?

marmijo: Done!

vars/kola.groovy
marmijo: Fixed!

jlebon (approved):
Nice, thanks!

PR #163, opened 2025-03-23
kolaTestIso: retry failed tests

dustymabe:
We've been seeing a lot of testiso flakes lately. Let's try to give ourselves a break by retrying like we do for normal kola runs.

This is slightly different because it's not done inside kola but outside kola. For the inside kola retry of testiso we'll wait for https://github.com/coreos/coreos-assembler/issues/3989 to happen.


marmijo (approved):
LGTM

aaradhak (approved):
+1 for merging this, we can work on a fix if this is throwing an issue.

PR #165, opened 2025-05-01
cosaPod: archive cosa git info and RPMs

jlebon:
The git info is already printed but having it be a proper archive makes it easier to access.

Also archive the full rpmdb package list.

### DIFF
--- a/vars/cosaPod.groovy
+++ b/vars/cosaPod.groovy
@@ -15,7 +15,11 @@ def call(params = [:], Closure body) {
     params['cmd'] = ["/usr/bin/dumb-init", "/usr/bin/sleep", "infinity"]
 
     pod(params) {
-        shwrap("cat /cosa/coreos-assembler-git.json || :")
+        shwrap("(cat /cosa/coreos-assembler-git.json || :) | tee coreos-assembler-git.json")
+        archiveArtifacts artifacts: "coreos-assembler-git.json"
+        shwrap("rpm -qa | sort > coreos-assembler-rpmdb.txt")
+        archiveArtifacts artifacts: "coreos-assembler-rpmdb.txt"
+        shwrap("rm coreos-assembler-git.json coreos-assembler-rpmdb.txt")
         body()
     }
 }


jlebon:
(Untested to be clear)

vars/cosaPod.groovy
```diff
@@ -15,7 +15,11 @@ def call(params = [:], Closure body) {
     params['cmd'] = ["/usr/bin/dumb-init", "/usr/bin/sleep", "infinity"]
 
     pod(params) {
-        shwrap("cat /cosa/coreos-assembler-git.json || :")
+        shwrap("(cat /cosa/coreos-assembler-git.json || :) | tee coreos-assembler-git.json")
+        archiveArtifacts artifacts: "coreos-assembler-git.json"
+        shwrap("rpm -qa | sort > coreos-assembler-rpmdb.txt")
+        archiveArtifacts artifacts: "coreos-assembler-rpmdb.txt"
+        shwrap("rm coreos-assembler-git.json coreos-assembler-rpmdb.txt")
```
dustymabe, line 22: I'm happy with this, but WDYT about including the list of RPMs inside the `coreos-assembler-git.json` from the beginning. Then it will get placed in our builds dir and uploaded too and we can even add support for displaying the diff to `cosa diff --source-control --from= --to=`.

jlebon: AFAIK we don't currently upload `coreos-assembler-git.json` separately but instead inlined as part of `meta.json`. Adding the rpm list there would make sense but probably not in `meta.json`. I'd probably just have `cmd-build` be the thing that creates that file in the builds dir from a fresh `rpm -qa | sort` query. I like how it plays into `cosa diff`.

Anyway, I guess even with the above, this is still useful to have I think to make it more easily accessible from the Jenkins interface itself.

dustymabe: > AFAIK we don't currently upload `coreos-assembler-git.json` separately 

Yeah. I guess we don't output that one to the build dir like we do: https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/42.20250410.3.2/x86_64/coreos-assembler-config-git.json

dustymabe (approved):
LGTM

jlebon:
OK, I didn't test this. But will try it out as soon as I merge it to verify that it works as expected.

PR #166, opened 2025-08-12
cosaBuild: call `cosa osbuild` separately

dustymabe:
In https://github.com/coreos/coreos-assembler/pull/4255 we are proposing to remove followup target build support from `cosa build`; i.e. it won't auto build the `qemu` target when doing a build.

Let's split this part out into a separate step here.


jlebon (approved):
WFM. It feels a bit like a contradiction of https://github.com/coreos/coreos-assembler/pull/4255, but coreos-ci-lib is about abstracting things out and in fact `cosaBuild()` already does lots of other things.

PR #168, opened 2026-05-04
kolaTestIso: check for testiso command before running tests

nikita-dubrovskii:
Part of folding `testiso` into `kola run`.

Issue: https://github.com/coreos/coreos-assembler/issues/3989

Related:
https://github.com/coreos/coreos-assembler/pull/4377
https://github.com/coreos/fedora-coreos-pipeline/pull/1339

### DIFF
--- a/vars/kolaTestIso.groovy
+++ b/vars/kolaTestIso.groovy
@@ -11,6 +11,11 @@ def call(params = [:]) {
     def extraArgs = params.get('extraArgs', "");
     def marker = params.get('marker', "");
 
+    // Check if kola has testiso command, skip if not available
+    if (shwrapRc("cosa kola help | grep -q testiso") != 0) {
+        echo "Skipping kola testiso: command not available in this version of kola"
+        return
+    }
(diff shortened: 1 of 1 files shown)


vars/kolaTestIso.groovy
```diff
@@ -11,6 +11,11 @@ def call(params = [:]) {
     def extraArgs = params.get('extraArgs', "");
     def marker = params.get('marker', "");
 
+    // Check if kola has testiso command, skip if not available
+    if (shwrapRc("cd ${cosaDir} && cosa kola help | grep -q testiso") != 0) {
```
dustymabe: do we need to `cd` for this?

dustymabe (approved):

vars/kolaTestIso.groovy
nikita-dubrovskii: Not really, that was a copy-paste. Now dropped.

dustymabe (approved):

PR #169, opened 2026-05-06
kola: add --inst-insecure flag to kola tests

nikita-dubrovskii:
Autodetection of dev builds fails on builds where 'dev' is not part of the filename scheme, like `rhcos-9.8.20260506-0-live-iso.x86_64.iso`. This causes kola tests to fail when they should run in insecure mode.

The obsolete `vars/kolaTestIso.groovy` used the insecure flag by default, so this restores that behavior explicitly.

### DIFF
--- a/vars/kola.groovy
+++ b/vars/kola.groovy
@@ -109,6 +109,11 @@ def call(params = [:]) {
             args += " --tag=!${tag}"
         }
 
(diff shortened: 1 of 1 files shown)


joelcapitao (approved):
LGTM

PR #170, opened 2026-05-06
cosaPod: add timeout to prevent pods from running indefinitely

aaradhak:
## Summary

- Adds a `timeout` parameter to `cosaPod` (default: 24 hours) to prevent hung or failed pods from blocking the pipeline indefinitely
- Callers can override the timeout (in hours) or disable it with `timeout: 0` for jobs like debug pods that need extended runtime
- Uses Jenkins' built-in `timeout` step which gracefully aborts the build on expiry

Fixes: https://github.com/coreos/fedora-coreos-pipeline/issues/1296

### DIFF
--- a/vars/cosaPod.groovy
+++ b/vars/cosaPod.groovy
@@ -1,6 +1,7 @@
 // Thin wrapper around pod() which auto-selects images and kvm.
 // Additional available parameters:
 //   image: string
+//   timeout: int (hours, default: 24, 0 to disable)
 def call(params = [:], Closure body) {
     if (params['image'] == null) {
         params['image'] = 'quay.io/coreos-assembler/coreos-assembler:main'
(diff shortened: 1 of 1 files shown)


vars/cosaPod.groovy
```diff
@@ -11,15 +12,26 @@ def call(params = [:], Closure body) {
         params['kvm'] = true
     }
 
+    def podTimeout = params.containsKey('timeout') ? params.remove('timeout') : 24
```
prestist, line 15: We could add a comment with an issue linked if we want to https://github.com/coreos/fedora-coreos-pipeline/issues/1296 for easy traceability. wdyt?

prestist (approved):
LGTM, you can ignore or do my comment very minor!

vars/cosaPod.groovy
aaradhak: The commit message has the link for the issue. I believe it should be good enough in this case may be. The code blame would point to the tracker issue.

PR #172, opened 2026-06-05
kola: split up kola runs only if necessary

dustymabe:
Now that cosa can schedule tests based on available resources [1] we will likely drop the `reprovision` tag from tests to enable them tests to all run together. Let's allow for both old and new ways of running things here.

[1] https://github.com/coreos/coreos-assembler/pull/4470

### DIFF
--- a/vars/kola.groovy
+++ b/vars/kola.groovy
@@ -149,15 +149,32 @@ def call(params = [:]) {
                     runKola(id, 'run', "--basic-qemu-scenarios ${skipSecureBootArg}")
                 }
             }
-            // normal run (without reprovision tests because those require a lot of memory)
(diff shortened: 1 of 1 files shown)


c4rt0 (approved):
At first the `jq -r...` didn't work locally as I attempted to verify if it successfully returns tests with 'reprovision' tag, but that turned out to be an issue with my local environment variables. 

If it comes to format:

```groovy
id = marker == "" ? "kola" : "kola-${marker}"
```
was honestly counterintuitive, after a bit of browsing it seems to be something like the python's equivalent: 

```python
  id = "kola" if marker == "" else f"kola-{marker}"
```
Learned something new here.

/lgtm

